PR #31, opened 2020-07-14
Select apprenticeships

MichaelYoung1981:


src/SFA.DAS.EmployerIncentives.Abstractions/Domain/AggregateRoot.cs
```diff
@@ -4,4 +4,8 @@ public abstract class AggregateRoot<IdType, EntityModel> : Entity<IdType, Entity
     {
         protected AggregateRoot(IdType id, EntityModel properties, bool isNew) : base(id, properties, isNew) { }        
     }
+
+    public abstract class AggregateRoot
```
Stephek, line 8: An aggregate root is an entity so it should have an identity.  This doesn't enforce that hence the object inheriting from it is probably a value object.  Value objects should contain most of the business logic so that would fit with NewApprenticeIncentive.

src/SFA.DAS.EmployerIncentives.Domain/NewApprenticeIncentive/NewApprenticeIncentive.cs
```diff
@@ -0,0 +1,21 @@
+﻿using System;
+using SFA.DAS.EmployerIncentives.Abstractions.Domain;
+using SFA.DAS.EmployerIncentives.ValueObjects;
+
+namespace SFA.DAS.EmployerIncentives.Domain.NewApprenticeIncentive
+{
+    public class NewApprenticeIncentive : AggregateRoot
```
Stephek, line 7: As mentioned above shouldn't this be a value object and inherit from SFA.DAS.EmployerIncentives.Abstractions.Domain.ValueObject?

src/SFA.DAS.EmployerIncentives.Domain/Services/NewApprenticeIncentiveEligibilityService.cs
```diff
@@ -0,0 +1,13 @@
+﻿using SFA.DAS.EmployerIncentives.Abstractions.Domain.Services;
+using SFA.DAS.EmployerIncentives.ValueObjects;
+
+namespace SFA.DAS.EmployerIncentives.Domain.Services
+{
+    public class NewApprenticeIncentiveEligibilityService : INewApprenticeIncentiveEligibilityService
+    {
+        public bool IsApprenticeshipEligible(Apprenticeship apprenticeship)
+        {
+            return NewApprenticeIncentive.NewApprenticeIncentive.IsApprenticeshipEligible(apprenticeship);
```
Stephek, line 10: Can we get read of the double NewApprenticeIncentive?  Maybe the namespace could be ApprenticeIncentive

src/SFA.DAS.EmployerIncentives.Domain/NewApprenticeIncentive/NewApprenticeIncentive.cs
```diff
@@ -0,0 +1,21 @@
...
+    {
+        private static readonly DateTime EligibilityStartDate = new DateTime(2020, 8, 1);
+
+        public static bool IsApprenticeshipEligible(Apprenticeship apprenticeship)
```
Stephek, line 11: It doesn't look right to have static methods on domain object since the domain object isn't acting on it's own data and is acting just as a service.  Any reason these rules can't be in the apprenticeship object?

src/SFA.DAS.EmployerIncentives.ValueObjects/Apprenticeship.cs
```diff
@@ -0,0 +1,18 @@
+﻿using System;
+
+namespace SFA.DAS.EmployerIncentives.ValueObjects
+{
+    public class Apprenticeship
```
Stephek, line 5: Can this inherit from ValueObject to enforce the immutability based on it's data?

src/SFA.DAS.EmployerIncentives.Abstractions/DTOs/ApprenticeshipDto.cs
```diff
@@ -0,0 +1,11 @@
+﻿using System;
+
+namespace SFA.DAS.EmployerIncentives.Abstractions.DTOs
+{
+    public class ApprenticeshipDto
+    {
+        public long UniqueLearnerNumber { get; set; }
```
Pauljgraham, line 7: We refer this as Uln in different places. Do we want to call it the same across the domain

src/SFA.DAS.EmployerIncentives.Abstractions/DTOs/ApprenticeshipDto.cs
```diff
@@ -0,0 +1,11 @@
...
+        public DateTime StartDate { get; set; }
+        public bool IsApproved { get; set; }
```
Pauljgraham, line 9: Not quite sure what the intent of the field IsApproved. Would it be better to say IsEligible

src/SFA.DAS.EmployerIncentives.Api/Controllers/EligibleApprenticeshipsQueryController.cs
```diff
@@ -0,0 +1,32 @@
...
+    [Route("eligible-apprenticeships")]
+    [ApiController]
+    public class EligibleApprenticeshipsQueryController : ApiQueryControllerBase
+    {
+        public EligibleApprenticeshipsQueryController(IQueryDispatcher queryDispatcher) : base(queryDispatcher)
+        {
+        }
+
+        [HttpGet("{uln}")]
+        public async Task<IActionResult> IsApprenticeshipEligible(long uln, [FromQuery]DateTime startDate, [FromQuery]bool isApproved)
```
Pauljgraham, line 19: My understanding is you may also need the expectedDate date, stop date, pause date and date of birth to determine if they are eligible

MichaelYoung1981: They're not part of this story so we can add them later on.

MichaelYoung1981:
> I don't believe the IsApproved flag is needed in your API. The apprentices will appear in the search if they are full apprenticeships

I've checked the code and it doesn't seem to be only for approved!

PR #44, opened 2020-07-29
Ei 162 create draft incentive application

MichaelYoung1981:


src/SFA.DAS.EmployerIncentives.Database/Tables/IncentiveApplication.sql
```diff
@@ -0,0 +1,10 @@
+﻿CREATE TABLE [dbo].[IncentiveApplication]
```
David-B-Read, line 1: Do these tables need indexing or is this just a write store?

MichaelYoung1981:
> A few comments. The concern I have is, we don't appear to re-validate that these apprentices are still eligible for the apprenticeship when we submit the application. I don't see a great deal of value adding it here but I always assumed as we save a state in DDD we ensure it's valid.
> 
> I'm still trying to get my head round separating the data storage from the domain layer as I've not seen this pattern before.

I see the re-validation being part of the submit process.

src/SFA.DAS.EmployerIncentives.Abstractions/DTOs/IncentiveClaimApprenticeshipDto.cs
```diff
@@ -0,0 +1,17 @@
+﻿using System;
+using SFA.DAS.Common.Domain.Types;
+
+namespace SFA.DAS.EmployerIncentives.Abstractions.DTOs
+{
+    public class IncentiveClaimApprenticeshipDto
+    {
+        public Guid Id { get; set; }
```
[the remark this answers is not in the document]
MichaelYoung1981: I was trying to keep id creation consistent but I don't mind creating them in the inner api.

src/SFA.DAS.EmployerIncentives.Api.AcceptanceTests/Steps/IncentiveApplicationCreatedSteps.cs
```diff
@@ -0,0 +1,58 @@
...
+            EmployerIncentiveApi.Response.StatusCode.Should().Be(_expectedResult);
+
+            using (var dbConnection = new SqlConnection(_testContext.SqlDatabase.DatabaseInfo.ConnectionString))
+            {
+                var application = await dbConnection.QueryAsync<IncentiveApplication>("SELECT * FROM IncentiveApplication WHERE Id = @IncentiveApplicationId",
+                    new { _request.IncentiveApplicationId });
+
+                application.Should().HaveCount(1);
+                application.Single().Id.Should().Be(_request.IncentiveApplicationId);
+            }
```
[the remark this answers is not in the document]
MichaelYoung1981: The data layer unit tests do that check.

src/SFA.DAS.EmployerIncentives.Abstractions/DTOs/IncentiveClaimApprenticeshipDto.cs
```diff
@@ -0,0 +1,17 @@
...
+        public int ApprenticeshipId { get; set; }
+        public string FirstName { get; set; }
+        public string LastName { get; set; }
+        public DateTime DateOfBirth { get; set; }
+        public long Uln { get; set; }
+        public DateTime PlannedStartDate { get; set; }
+        public ApprenticeshipEmployerType ApprenticeshipEmployerTypeOnApproval { get; set; }
+    }
+}
```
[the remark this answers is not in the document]
MichaelYoung1981: I've assumed the same.  If we do need it after we can add it in.

src/SFA.DAS.EmployerIncentives.Data/Map/DataExtensions.cs
```diff
@@ -35,5 +38,36 @@ public static AccountModel MapSingle(this IEnumerable<Models.Account> accounts)
...
         }
+
+        internal static Models.IncentiveApplication Map(this IncentiveApplicationModel model)
+        {
+            return new IncentiveApplication
+            {
+                Id = model.Id,
+                Status = model.Status,
+                DateSubmitted = model.DateSubmitted,
+                SubmittedBy = model.SubmittedBy,
```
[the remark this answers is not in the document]
MichaelYoung1981: We aren't doing submission yet. 😄

src/SFA.DAS.EmployerIncentives.Database/Tables/IncentiveApplication.sql
MichaelYoung1981: No read yet.  There will be at some point but we can add the index then.

src/SFA.DAS.EmployerIncentives.Api.AcceptanceTests/Features/IncentiveApplicationCreated.feature
```diff
@@ -0,0 +1,11 @@
+﻿@database
+@api
+Feature: IncentiveApplicationCreated
+	In order to select apprenticeships to apply for an incentive for
+	As an employer
+	I want to know if an apprenticeship is eligible
+
+Scenario: Incentive Application is created
+	Given I am applying for the New Apprenticeship Incentive
```
Stephek, line 9: It's generally better to write in the third person because it's not clear who or what is "I"?  Is it an employer, and apprentice a provider a support user?  So better to use An employer applying for ...

src/SFA.DAS.EmployerIncentives.Domain/Factories/IncentiveApplicationFactory.cs
```diff
@@ -0,0 +1,31 @@
...
+    public class IncentiveApplicationFactory : IIncentiveApplicationFactory
+    {
+        public IncentiveApplication CreateNew(Guid id, long accountId, long accountLegalEntityId)
+        {
+            return IncentiveApplication.New(id, accountId, accountLegalEntityId);
+        }
+
+        public Apprenticeship CreateNewApprenticeship(Guid id, int apprenticeshipId, string firstName, string lastName, DateTime dateOfBirth, long uln, DateTime plannedStartDate, ApprenticeshipEmployerType apprenticeshipEmployerTypeOnApproval)
+        {
+            var model = new ApprenticeshipModel
```
Stephek, line 17: I don't think we should be creating entities this way.  The Create(model) is for hydrating from a data store.  Otherwise any domain rules could be bypassed.  So this should probably just call Create(id, FirstName, LastName, othermandatoryfields etc...)  Then the application layer (the use case) can call other methods like apprenticeship.SetPlannedStartDate(date) etc.... Then That method could check that he date is not in the past etc...  Create should just create on object the minimum required fields.

src/tests/SFA.DAS.EmployerIncentives.Api.UnitTests/Application/WhenCreatingAnApplication.cs
```diff
@@ -0,0 +1,68 @@
...
+            _fixture = new Fixture();
+            _sut = new ApplicationCommandController(_mockCommandDispatcher.Object);
+
+            _mockCommandDispatcher
+                .Setup(m => m.Send(It.IsAny<AddLegalEntityCommand>(), It.IsAny<CancellationToken>()))
+                .Returns(Task.CompletedTask);
+        }
+
+        [Test]
+        public async Task Then_a_AddLegalEntityCommand_command_is_dispatched()
```
Stephek, line 36: Cut and paste name issue.

src/SFA.DAS.EmployerIncentives.Domain/Factories/IncentiveApplicationFactory.cs
MichaelYoung1981: I agree but in our case they're all minimum required fields, without any of them the object is invalid in our context.

Stephek: So should all be passed in when creating the object.

PR #46, opened 2020-07-31
Ei 87 update application status

David-B-Read:


src/SFA.DAS.EmployerIncentives.Api.AcceptanceTests/Features/IncentiveClaimSubmitted.feature
```diff
@@ -0,0 +1,23 @@
...
+	Given an employer has entered incentive claim application details
+	When the invalid application id is submittted
+	Then the application status is not updated
+	And the service responds with an error
+
+Scenario: Incentive Application is submitted with invalid account id
+	Given an employer has entered incentive claim application details
+	When the invalid account id is submittted
+	Then the application status is not updated
+	And the service responds with an error
```
MichaelYoung1981, line 23: Personally, I think these are too granular to be acceptance tests.

src/SFA.DAS.EmployerIncentives.Api/Controllers/ApplicationCommandController.cs
```diff
@@ -21,5 +23,20 @@ public async Task<IActionResult> CreateIncentiveApplication([FromBody]CreateInce
             await SendCommandAsync(new CreateIncentiveApplicationCommand(request.IncentiveApplicationId, request.AccountId, request.AccountLegalEntityId, request.Apprenticeships));
             return Created($"/applications/{request.IncentiveApplicationId}", null);
         }
+
+        [HttpPost("/submit-application")]
```
MichaelYoung1981, line 27: That isn't a restful url...don't ask me what the correct url would be tho! 😉 
Maybe something like patch /applications/{id}/

src/SFA.DAS.EmployerIncentives.Commands/SubmitIncentiveApplication/SubmitIncentiveApplicationCommandValidator.cs
```diff
@@ -0,0 +1,26 @@
...
+            if (item.IncentiveApplicationId == default)
+            {
+                result.AddError("IncentiveApplicationId", "Is not set");
+            }
+
+            if (item.AccountId == default)
+            {
+                result.AddError("AccountId", "Is not set");
+            }
+
```
MichaelYoung1981, line 22: Do the other properties need validating?

src/SFA.DAS.EmployerIncentives.Data/IncentiveApplicationDataRepository.cs
```diff
@@ -19,5 +22,26 @@ public async Task Add(IncentiveApplicationModel incentiveApplication)
             await _dbContext.AddAsync(incentiveApplication.Map());
             await _dbContext.SaveChangesAsync();
         }
+
+        public async Task<IncentiveApplicationModel> Get(Guid incentiveApplicationId)
+        {
+            var incentiveApplication = _dbContext.Applications.AsNoTracking().FirstOrDefault(a => a.Id == incentiveApplicationId);
```
MichaelYoung1981, line 28: If we're not tracking and have to update the object in the below method aren't we missing out on one of the bigger benefits of EF?

src/SFA.DAS.EmployerIncentives.Commands/SubmitIncentiveApplication/SubmitIncentiveApplicationCommandHandler.cs
```diff
@@ -0,0 +1,37 @@
...
+            var application = await _domainRepository.Find(command.IncentiveApplicationId);
+
+            if (application == null || application.AccountId != command.AccountId)
+            {
+                throw new InvalidRequestException();
+            }
+
+            application.Submit(command.DateSubmitted, command.SubmittedBy);
+
+            await _domainRepository.Save(application);
```
MichaelYoung1981, line 34: We'll need to push a submitted event out of this method.

src/SFA.DAS.EmployerIncentives.Data/IncentiveApplicationDataRepository.cs
David-B-Read: I had to remove tracking so that the unit tests for update would work, happy to take suggestions on re-writing the tests to work with tracking

David-B-Read: I removed tracking because the unit tests were throwing the infamous "The instance of entity type 'Item' cannot be tracked because another instance with the same key value for {'Id'} is already being tracked" exception. A rewrite of the update code seems to have solved this.

MichaelYoung1981:
Still needs the submit event pushing but other than that it looks good to me.

src/SFA.DAS.EmployerIncentives.Commands/SubmitIncentiveApplication/SubmitIncentiveApplicationCommandHandler.cs
```diff
@@ -0,0 +1,51 @@
...
+            var events = new List<EmployerIncentiveClaimSubmittedEvent>
+            {
+                new EmployerIncentiveClaimSubmittedEvent
+                {
+                    AccountId = command.AccountId,
+                    IncentiveClaimApplicationId = command.IncentiveApplicationId
+                }
+            };
+
+            await _eventPublisher.Publish(events);
```
David-B-Read, line 48: @MichaelYoung1981 Is this sufficient for publishing the event or is other wiring up required?

MichaelYoung1981: Was that already in there and I missed it? 🤣   That should be fine.

PR #56, opened 2020-08-06
EI-203 Apply - update draft application with a new set of apprenticeships

ruslanss:


src/SFA.DAS.EmployerIncentives.Api.AcceptanceTests/SFA.DAS.EmployerIncentives.Api.AcceptanceTests.csproj
```diff
@@ -1,4 +1,4 @@
-﻿<Project Sdk="Microsoft.NET.Sdk.Web">
+﻿<Project Sdk="Microsoft.NET.Sdk">
```
ruslanss, line 1: this fixes SpecFlow extensions with all tests running, passing and able to debug

src/SFA.DAS.EmployerIncentives.Data/IncentiveApplication/IncentiveApplicationDataRepository.cs
```diff
@@ -39,8 +43,28 @@ public async Task Update(IncentiveApplicationModel incentiveApplication)
             if (existingApplication != null)
             {
                 _dbContext.Entry(existingApplication).CurrentValues.SetValues(model);
+                foreach (var apprenticeship in RemovedApprenticeships(existingApplication, model))
```
ruslanss, line 46: still having to do this in EF Core makes me upset 😢

src/SFA.DAS.EmployerIncentives.Commands/UpdateIncentiveApplication/UpdateIncentiveApplicationCommandHandler.cs
```diff
@@ -0,0 +1,56 @@
...
+            _domainRepository = domainRepository;
+        }
+
+        public async Task Handle(UpdateIncentiveApplicationCommand command, CancellationToken cancellationToken = default)
+        {
+            var application = await _domainRepository.Find(command.IncentiveApplicationId);
+
+            foreach (var apprenticeship in RemovedApprenticeships(command, application))
+            {
+                application.RemoveApprenticeship(apprenticeship);
```
Stephek, line 29: Couldn't you just have a method on the domain object like

application.SetApprenticeships(IEnumerable<Apprenticeship> apprenticeships)

where Apprenticeship is a value object and let the domain object decide whether to add or not since there may be additional rules it may want to check.  Aren't all the saves apprenticeships being replaced anyway?

src/SFA.DAS.EmployerIncentives.Data/IncentiveApplication/IncentiveApplicationDataRepository.cs
Stephek: Why removing them?  Are they not all being replaced?  It's just a case of saving the state of the model so you could just delete everything and replace it since you are just hydrating the model to the datastore for persistence.

src/SFA.DAS.EmployerIncentives.Domain/IncentiveApplications/IncentiveApplication.cs
```diff
@@ -39,11 +44,18 @@ public void AddApprenticeship(Apprenticeship apprenticeship)
             Model.ApprenticeshipModels.Add(apprenticeship.GetModel());
         }
 
+        public void RemoveApprenticeship(Apprenticeship apprenticeship)
```
Stephek, line 47: Can we not just set the required apprenticeships in the application rather than have add and remove methods?  As an analogy if the application had a name you wouldn't have a RemoveName and AddName method.  You'd just set the name so it would be more simple to just set the apprenticeships

src/SFA.DAS.EmployerIncentives.Api.AcceptanceTests/SFA.DAS.EmployerIncentives.Api.AcceptanceTests.csproj
MichaelYoung1981: When I changed this it broke some other stuff!

PR #62, opened 2020-08-13
Ei 144 check uln has not been included in prev application

Pauljgraham:


src/SFA.DAS.EmployerIncentives.Data/Apprenticeship/UlnQueryRepository.cs
```diff
@@ -0,0 +1,29 @@
+﻿using System.Linq;
+using System.Threading.Tasks;
+using Microsoft.EntityFrameworkCore;
+using SFA.DAS.EmployerIncentives.Data.Models;
+using SFA.DAS.EmployerIncentives.Enums;
+
+namespace SFA.DAS.EmployerIncentives.Data.Apprenticeship
+{
+    public class UlnQueryRepository : IUlnQueryRepository
```
MichaelYoung1981, line 9: I don't think we should be using QueryRepos for domain logic. In reality, this probably just means renaming this class.

src/SFA.DAS.EmployerIncentives.Queries/NewApprenticeIncentive/GetApprenticeshipEligibility/GetApprenticeshipEligibilityQueryHandler.cs
```diff
@@ -1,25 +1,29 @@
...
+            _queryRepository = queryRepository;
         }
 
         public async Task<GetApprenticeshipEligibilityResponse> Handle(GetApprenticeshipEligibilityRequest query, CancellationToken cancellationToken = default)
         {
             var apprenticeship = query.Apprenticeship.ToApprenticeship();
+            var ulnUsed = await _queryRepository.UlnAlreadyOnSubmittedIncentiveApplication(query.Apprenticeship.UniqueLearnerNumber);
 
-            var isEligible = _eligibilityService.IsApprenticeshipEligible(apprenticeship);
+            var isEligible = !ulnUsed && _eligibilityService.IsApprenticeshipEligible(apprenticeship);
```
MichaelYoung1981, line 26: I'm wondering if it'd be worth moving all of the logic to the eligibility service.  That'd mean having a repo method to get any existing claimapprenticeship records for the uln and passing them to the service and allowing it to decide what to do with them.

src/SFA.DAS.EmployerIncentives.Data/Apprenticeship/UlnQueryRepository.cs
Pauljgraham: Happy to call it something else. Just not sure what.

src/SFA.DAS.EmployerIncentives.Queries/NewApprenticeIncentive/GetApprenticeshipEligibility/GetApprenticeshipEligibilityQueryHandler.cs
Pauljgraham: That's an interesting point. I wanted to do that originally, but the UlnQueryRepo class needs the Db Context which exists in Data project. We can do it by putting the interface into the abstractions layer and let the IoC resolve the class. 

Happy to do it this way if you want to. We could call the interface/class IUlnValidator

src/SFA.DAS.EmployerIncentives.Data/Apprenticeship/UlnQueryRepository.cs
Pauljgraham: renamed to UlnValidationServce

src/SFA.DAS.EmployerIncentives.Queries/NewApprenticeIncentive/GetApprenticeshipEligibility/GetApprenticeshipEligibilityQueryHandler.cs
Pauljgraham: done

PR #74, opened 2020-08-26
EI-296 Fix-  Stopped messages getting published twice

Pauljgraham:
Removed the MultiEventPublisher and replaced it with IEventPublisher


David-B-Read (approved):

MichaelYoung1981 (changes requested):
We shouldn't really be using IEventPublisher either as it doesn't include the message in the Unit of Work.  There is a UnitOfWorkContext class (I think!) that we should use instead so the messages don't get published if the transaction fails.

Stephek:
> We shouldn't really be using IEventPublisher either as it doesn't include the message in the Unit of Work. There is a UnitOfWorkContext class (I think!) that we should use instead so the messages don't get published if the transaction fails.

IEventPublisher just adds messages to an in memory collection which is then used by the UOW to publish the messages.  It makes it more intuitive that you are publishing an event I guess.

MichaelYoung1981 (approved):

PR #89, opened 2020-09-21
Dasd 8156 convert sql connection to managed

NickGraham101:


NickGraham101:
Work moved to quality guild

PR #91, opened 2020-09-23
EI-279 Update VRF Case Status for Legal Entity

ruslanss:


src/SFA.DAS.EmployerIncentives.Database/Scripts/Script.PreDeployment1.sql
```diff
@@ -0,0 +1,11 @@
+﻿/*
```
David-B-Read, line 1: Is this file needed?

src/SFA.DAS.EmployerIncentives.Api.AcceptanceTests/Features/UpdateVrfCaseDetailsForNewApplications.feature
```diff
@@ -1,4 +1,5 @@
-﻿@database
+﻿@ignore
```
David-B-Read, line 1: Is the ignore set here because the tests won't run on CI?

David-B-Read (approved):

src/SFA.DAS.EmployerIncentives.Api.AcceptanceTests/DataAccess.cs
```diff
@@ -23,12 +23,21 @@ public Account GetAccountByLegalEntityId(long legalEntityId)
...
+        public Account GetAccountByAccountLegalEntityId(long accountId, long accountLegalEntityId)
+        {
+            using var dbConnection = new SqlConnection(_connectionString);
+            var account = dbConnection.QuerySingle<Account>("SELECT * FROM Accounts WHERE Id = @accountId AND AccountLegalEntityId = @AccountLegalEntityId",
+                new { accountId, accountLegalEntityId });
+
+            return account;
+        }
+
         public void SetupAccount(Account account)
```
Stephek, line 35: Can you use Dapper contrib as in EI-49?  It's a lot simpler I think.

Stephek (changes requested):

src/SFA.DAS.EmployerIncentives.Api.AcceptanceTests/DataAccess.cs
ruslanss: hmm this is a test code , can refactor once EI-49 is merged?

src/SFA.DAS.EmployerIncentives.Api.AcceptanceTests/Steps/UpdateVrfCaseStatusForLegalEntitySteps.cs
```diff
@@ -0,0 +1,67 @@
+﻿using FluentAssertions;
+using SFA.DAS.EmployerIncentives.Data.Models;
+using System;
+using System.Net;
+using System.Threading.Tasks;
+using TechTalk.SpecFlow;
+
+namespace SFA.DAS.EmployerIncentives.Api.AcceptanceTests.Steps
+{
+    [Binding]
```
Stephek, line 10: Please add a 

[Scope(Feature = "... ")]  attribute

src/SFA.DAS.EmployerIncentives.Api/Controllers/AccountCommandController.cs
```diff
@@ -37,5 +38,12 @@ public async Task<IActionResult> SignAgreement([FromRoute] long accountId, [From
             await SendCommandAsync(new SignLegalEntityAgreementCommand(accountId, accountLegalEntityId, request.AgreementVersion));
             return NoContent();
         }
+
+        [HttpPatch("/legalentities/{hashedLegalEntityId}/vendorregistrationform/status")]
```
Stephek, line 42: We shouldn't have path like this.  The path locates the resource

/legalentities/{hashedLegalEntityId}

the patched data is in the body.  So it's a REST interface rather than a RPC call.

The body should details what has changed e.g. as per https://en.wikipedia.org/wiki/JSON_Patch

So in this example would be 

[
    { "op": "replace", "path": "/vendorregistrationform/status", "value": "newStatus" }
]

Stephek (changes requested):

src/SFA.DAS.EmployerIncentives.Commands/Persistence/IAccountDomainRepository.cs
```diff
@@ -1,9 +1,13 @@
 ﻿using SFA.DAS.EmployerIncentives.Abstractions.Domain;
 using SFA.DAS.EmployerIncentives.Domain.Accounts;
+using System.Collections.Generic;
+using System.Threading.Tasks;
 
 namespace SFA.DAS.EmployerIncentives.Commands.Persistence
 {
     public interface IAccountDomainRepository : IDomainRepository<long, Account>
     {
+        Task<IEnumerable<Account>> GetByLegalEntityId(long legalEntityId);
```
Stephek, line 10: These shouldn't be on the domain respository.  Use the IQueryRepository to get data then look up the domain entity using the IDomainRepository.Find (by entity Id).

The entity Id is the key to the domain model.  It's the logical write model it's not designed for querying.

Stephek (changes requested):

src/SFA.DAS.EmployerIncentives.Commands/Persistence/AccountDomainRepository.cs
```diff
@@ -24,12 +26,26 @@ public async Task<Account> Find(long id)
...
             if (aggregate.IsNew)
-            {               
+            {
                 return _accountDataRepository.Add(aggregate.GetModel());
             }
 
             return _accountDataRepository.Update(aggregate.GetModel());
         }
+
+        public async Task<IEnumerable<Account>> GetByLegalEntityId(long legalEntityId)
```
Stephek, line 36: Please remove form the Domain repository. Use the existing query repository to find the account Id by other fields.  Then get the entity from the domain repository,  Treat them as separate read/write models with separate concerns.

src/SFA.DAS.EmployerIncentives.Commands/UpsertLegalEntity/UpsertLegalEntityCommandHandler.cs
```diff
@@ -0,0 +1,45 @@
...
+
+        public UpsertLegalEntityCommandHandler(IAccountDomainRepository domainRepository, IHashingService hashingService)
+        {
+            _domainRepository = domainRepository;
+            _hashingService = hashingService;
+        }
+
+        public async Task Handle(UpsertLegalEntityCommand command, CancellationToken cancellationToken = default)
+        {
+            var account = await _domainRepository.Find(command.AccountId);
```
Stephek, line 23: Would be better if this published the AddCommand or updatedCommand depending on if the account exists.  The work don in EI-49 would make this easier to implement.

src/SFA.DAS.EmployerIncentives.Commands/Persistence/IAccountDomainRepository.cs
MichaelYoung1981: For what benefit?  It'll slow down the process and bloat the code.

src/SFA.DAS.EmployerIncentives.Commands/Persistence/AccountDomainRepository.cs
MichaelYoung1981: For what benefit?  It'll slow down the process and bloat the code.  Not being able to get a domain object by anything other than the primary key is a step too far imo.

src/SFA.DAS.EmployerIncentives.Api/Controllers/AccountCommandController.cs
MichaelYoung1981: This is an interesting one, you're definitely right about it not being RESTful but what would the implementation of a generic patch method look like?  Is it worth taking this as tech debt with a spike to see how we'd implement the patch method without it becoming a bag of spanners?

Stephek: Yes a tech debt spike would be useful as the patch approach should be applied elsewhere in a consistent manor.

src/SFA.DAS.EmployerIncentives.Commands/Persistence/AccountDomainRepository.cs
Stephek: Because the domain repository is purely for hydrating and dehydrating domain objects form the logical write model.  Querying data for other purposes should be done through the existing query interface so that interface can be used elsewhere and indexed appropriately.  It keeps good separation of concerns and prevents the code from becoming a big mess.  It's a pattern following CQRS and although we don't have physical read/write models taking this approach removes the logical distinction.  However as this is supposedly a throw away project then if you want to add querying to this interface then I will resolve this.

MichaelYoung1981: It's not querying per-se tho, it's merely retrieval of a domain object to update isn't it?

src/SFA.DAS.EmployerIncentives.Commands/Persistence/IAccountDomainRepository.cs
Stephek: Because this interface is just for hydrating and persisting a single domain object from the logical write store.

There is already an existing IQueryRepository to retrieve data so that should be used as that is it's concern then publish the commands to do the work on each individual domain object.  That would be more performant since it will could handle multiple domain objects in parallel each observing he existing locking strategy.

By adding a query to this interface encourages batch processing.

src/SFA.DAS.EmployerIncentives.Domain/Accounts/LegalEntity.cs
```diff
@@ -38,5 +55,25 @@ public void SignedAgreement(int signedAgreementVersion, int minimumRequiredVersi
...
         }
+
+        public void SetHashedLegalEntityId(string hashedId)
+        {
+            Model.HashedLegalEntityId = hashedId;
+        }
+
+        internal void UpdateVendorRegistrationCaseStatus(string caseId, string vendorId, string status, DateTime caseStatusLastUpdatedDate)
+        {
+            if (VrfStatusIsCompleted()) return;
```
Pauljgraham, line 66: This logic looks wrong to me. I understood that we are supposed to ignore the incoming update if the new status is "Case request complete" . This logic updates the record with the status "Case request complete" and the all subsequent updates are then ignored.

src/tests/SFA.DAS.EmployerIncentives.Commands.UnitTests/UpdateVrfCaseStatusForLegalEntity/Handlers/WhenHandlingUpdateVrfCaseDetailsForLegalEntityCommand.cs
```diff
@@ -0,0 +1,110 @@
...
+            accounts[0].AddLegalEntity(1, legalEntityToBeUpdated);
+            accounts[0].AddLegalEntity(2, _fixture.Create<LegalEntity>());
+            accounts[0].AddLegalEntity(3, _fixture.Create<LegalEntity>());
+
+            accounts[1].AddLegalEntity(4, _fixture.Create<LegalEntity>());
+            accounts[1].AddLegalEntity(5, legalEntityToBeUpdated);
+
+            _mockDomainRepository.Setup(x => x.GetByHashedLegalEntityId(caseCompletedCommand.HashedLegalEntityId)).ReturnsAsync(accounts);
+            await _sut.Handle(caseCompletedCommand);
+
```
Pauljgraham, line 91: This is proving that after the status update to "Case request complete" we can no longer update the record. I think this is incorrect.

Pauljgraham (changes requested):

src/SFA.DAS.EmployerIncentives.Domain/Accounts/LegalEntity.cs
MichaelYoung1981: The logic is correct, it should set it to complete once and then ignore all future updates.

ruslanss: Yes, however we agreed to remove VendorID from the status update. My bad i shouldn't have added it in the first place.

Stephek (approved):

Pauljgraham (approved):

PR #94, opened 2020-09-25
Ei 49 domain events and uow fix

Pauljgraham:
This is the PR for calculating the Apprenticeship Incentives


src/SFA.DAS.EmployerIncentives.Api.AcceptanceTests/Features/ApprenticeshipIncentiveCreated.feature
```diff
@@ -0,0 +1,27 @@
+﻿@database
```
David-B-Read, line 1: A few typos here (apprentiveship, incentice)

src/SFA.DAS.EmployerIncentives.Database/Tables/ApprenticeshipIncentive.sql
```diff
@@ -0,0 +1,16 @@
+﻿CREATE TABLE [dbo].ApprenticeshipIncentive
```
David-B-Read, line 1: The majority of this data is already stored in the IncentiveApplicationApprenticeship table, is there a reason for duplicating?

src/SFA.DAS.EmployerIncentives.Commands.Types/ApprenticeshipIncentive/CreateCommand.cs
```diff
@@ -0,0 +1,37 @@
+﻿using SFA.DAS.EmployerIncentives.Abstractions.Commands;
+using SFA.DAS.EmployerIncentives.Abstractions.Logging;
+using SFA.DAS.EmployerIncentives.Domain.Accounts;
+using SFA.DAS.EmployerIncentives.Infrastructure.DistributedLock;
+using System;
+
+namespace SFA.DAS.EmployerIncentives.Commands.Types.ApprenticeshipIncentive
+{
+    public class CreateCommand : ICommand, ILockIdentifier, ILogWriter
```
MichaelYoung1981, line 9: Create what?

src/SFA.DAS.EmployerIncentives.Commands/ApprenticeshipIncentive/CalculateEarnings/CalculateEarningsCommandValidator.cs
```diff
@@ -0,0 +1,31 @@
+﻿using System.Threading.Tasks;
+using SFA.DAS.EmployerIncentives.Abstractions.Commands;
+using SFA.DAS.EmployerIncentives.Commands.Types.ApprenticeshipIncentive;
+
+namespace SFA.DAS.EmployerIncentives.Commands.ApprenticeshipIncentive.CalculateEarnings
+{
+    public class CalculateEarningsCommandValidator : IValidator<CalculateEarningsCommand>
```
MichaelYoung1981, line 7: Are these validators actually necessary?  We're protecting from ourselves as these will never be based on user input.

src/SFA.DAS.EmployerIncentives.Commands/ApprenticeshipIncentive/Create/CreateCommandHandler.cs
```diff
@@ -0,0 +1,53 @@
...
+namespace SFA.DAS.EmployerIncentives.Commands.ApprenticeshipIncentive.Create
+{
+    public class CreateCommandHandler : ICommandHandler<CreateCommand>
+    {
+        private readonly IIncentiveApplicationDomainRepository _applicationDomainRepository;
+        private readonly IApprenticeshipIncentiveFactory _apprenticeshipIncentiveFactory;
+        private readonly IApprenticeshipIncentiveDomainRepository _apprenticeshipIncentiveDomainRepository;
+        
+
+        public CreateCommandHandler(
```
MichaelYoung1981, line 19: Given we're duplicating data because of the different bounded context I think we need to have a greater distinction between bounded contexts (maybe a new db, new projects etc) as it's going to be difficult to understand what each data item is for/

src/SFA.DAS.EmployerIncentives.Commands/ApprenticeshipIncentive/Create/CreateCommandHandler.cs
```diff
@@ -0,0 +1,53 @@
...
+                    apprenticeship.Id,
+                    new Account(application.AccountId),
+                    new Apprenticeship(
+                        apprenticeship.ApprenticeshipId,
+                        apprenticeship.FirstName,
+                        apprenticeship.LastName,
+                        apprenticeship.DateOfBirth,
+                        apprenticeship.Uln,
+                        apprenticeship.ApprenticeshipEmployerTypeOnApproval
+                    ),
```
MichaelYoung1981, line 46: Do we know that we need all of these fields?

src/SFA.DAS.EmployerIncentives.Commands/Persistence/ApprenticeshipIncentiveDomainRepository.cs
```diff
@@ -0,0 +1,53 @@
...
+
+        public async Task Save(Domain.ApprenticeshipIncentives.ApprenticeshipIncentive aggregate)
+        {
+            if (aggregate.IsNew)
+            {
+                await _apprenticeshipIncentiveDataRepository.Add(aggregate.GetModel());
+            }
+            else
+            {
+                await _apprenticeshipIncentiveDataRepository.Update(aggregate.GetModel());
```
MichaelYoung1981, line 44: Should this set IsNew to false to it doesn't call add again if Save is called twice?

src/SFA.DAS.EmployerIncentives.Data/ApprenticeshipIncentives/ApprenticeshipIncentiveDataRepository.cs.cs
```diff
@@ -0,0 +1,51 @@
+﻿using Microsoft.EntityFrameworkCore;
```
MichaelYoung1981, line 1: File name has a .cs.cs extension

src/SFA.DAS.EmployerIncentives.Database/Tables/ApprenticeshipIncentive.sql
MichaelYoung1981: @David-B-Read It's a different bounded context (which needs to be made clearer)

MichaelYoung1981 (dismissed):

src/SFA.DAS.EmployerIncentives.Commands.Types/ApprenticeshipIncentive/CreateCommand.cs
Stephek: It's in the ApprenticeshipIncentive namespace so create an ApprenticeshipIncentive.  Does it need to be more wordy?

src/SFA.DAS.EmployerIncentives.Commands/ApprenticeshipIncentive/CalculateEarnings/CalculateEarningsCommandValidator.cs
Stephek: Just following the pattern.  The command handler doesn't know how or who entered the command data so is just protecting it's invariants.

src/SFA.DAS.EmployerIncentives.Commands/ApprenticeshipIncentive/Create/CreateCommandHandler.cs
Stephek: Yep.  We put the new bounded context into a separate namespace/folder but separate db and projects sounds like a better approach.

src/SFA.DAS.EmployerIncentives.Commands.Types/ApprenticeshipIncentive/CreateCommand.cs
MichaelYoung1981: I think so as it isn't immediately obvious what it does.

src/SFA.DAS.EmployerIncentives.Commands/ApprenticeshipIncentive/Create/CreateCommandHandler.cs
MichaelYoung1981: I've created a tech debt ticket (https://skillsfundingagency.atlassian.net/browse/EI-417) to split out the projects but @Pauljgraham is going to create a schema in the db for the tables as a starting point

src/SFA.DAS.EmployerIncentives.Commands/ApprenticeshipIncentive/Create/CreateCommandHandler.cs
Stephek: They define the apprenticeship.  We definitely need some to calculate the earnings.  FirstName and LastName are there so the UI can display without having to reference the application.

src/SFA.DAS.EmployerIncentives.Commands/ApprenticeshipIncentive/CalculateEarnings/CalculateEarningsCommandValidator.cs
MichaelYoung1981: Just seems like unnecessary code given these are handling domain events from the same solution.

src/SFA.DAS.EmployerIncentives.Commands/Persistence/ApprenticeshipIncentiveDomainRepository.cs
Stephek: The distributed lock should prevent this from being called twice.  We don't persist the isNew state of the domain object, it's isNew true when created else false if the state is loaded from the db.  It's a good point but if called twice they would have  two different Id's anyway.

src/SFA.DAS.EmployerIncentives.Commands/ApprenticeshipIncentive/Create/CreateCommandHandler.cs
Stephek: We did wonder about using a different schema in the db rather than dbo.  But that doesn't seem to be common practice here.

src/SFA.DAS.EmployerIncentives.Api.AcceptanceTests/Features/ApprenticeshipIncentiveCreated.feature
Stephek: Fixed

src/SFA.DAS.EmployerIncentives.Commands.Types/ApprenticeshipIncentive/CreateCommand.cs
Stephek: Done

src/SFA.DAS.EmployerIncentives.Commands/ApprenticeshipIncentive/CalculateEarnings/CalculateEarningsCommandValidator.cs
Stephek: Updated to not be required for DomainCommands.

src/SFA.DAS.EmployerIncentives.Database/Tables/ApprenticeshipIncentive.sql
```diff
@@ -0,0 +1,16 @@
...
+	[FirstName] NVARCHAR(100) NOT NULL,
+	[LastName] NVARCHAR(100) NOT NULL,
+	[DateOfBirth] DATETIME2 NOT NULL,
+	[ULN] BIGINT NOT NULL,	
+	[EmployerType] INT NOT NULL,
+	[PlannedStartDate] DATETIME2 NOT NULL,
+	[IncentiveApplicationApprenticeshipId] UNIQUEIDENTIFIER NOT NULL
+)
+GO
+CREATE UNIQUE CLUSTERED INDEX IX_ApprenticeshipIncentive ON [incentives].[ApprenticeshipIncentive] (AccountId, ApprenticeshipId)
```
Pauljgraham, line 15: Actually this unique index doesn't look quite right. The rule we are imposing is a student cannot be funded if he has previously been funded. So I think the unique index should be on the ULN field.

Pauljgraham (commented):
A concern about the unique index

src/SFA.DAS.EmployerIncentives.Database/Tables/ApprenticeshipIncentive.sql
MichaelYoung1981: Can't be ULN but could be account and ULN...if the learner changes employer before an incentive payment has been made then the new employer can apply.  That rule is currently being reviewed so not sure what the outcome will be.

Stephek: The business rules should be enforced in the code.  Since they may change.  The unique index is because the ApprenticeshipId, AccountId combination should be unique in this table.  Is that not true?

SreekanthBadigenchula:
Tests are all GREEN on this PR

David-B-Read (approved):

PR #95, opened 2020-09-30
Ei 362 handle rejected case statuses

David-B-Read:


src/SFA.DAS.EmployerIncentives.Domain/Accounts/LegalEntityVrfCaseStatus.cs
```diff
@@ -0,0 +1,10 @@
+﻿
+namespace SFA.DAS.EmployerIncentives.Domain.Accounts
+{
+    public static class LegalEntityVrfCaseStatus
```
Stephek, line 4: Do we need the LegalEntity bit in the name?

David-B-Read: It's the VRF case status for a given legal entity so thought that naming would be an indication of scope

src/SFA.DAS.EmployerIncentives.Data/IncentiveApplication/IncentiveApplicationQueryRepository.cs
```diff
@@ -66,6 +67,14 @@ private static IncentiveApplicationApprenticeshipDto MapToApprenticeshipDto(Ince
...
         private static bool MapBankDetailsRequired(string vrfCaseStatus)
         {
+            if (vrfCaseStatus != null && 
+                  (vrfCaseStatus.Equals(LegalEntityVrfCaseStatus.RejectedDataValidation, StringComparison.InvariantCultureIgnoreCase)
+                || vrfCaseStatus.Equals(LegalEntityVrfCaseStatus.RejectedVer1, StringComparison.InvariantCultureIgnoreCase)
+                || vrfCaseStatus.Equals(LegalEntityVrfCaseStatus.RejectedVerification, StringComparison.InvariantCultureIgnoreCase)))
+            {
+                return true;
+            }
+
```
Pauljgraham, line 77: would it read a little clearer if we just returned

return (String.IsNullOrWhiteSpace(vrfCaseStatus) ||                (vrfCaseStatus.Equals(LegalEntityVrfCaseStatus.RejectedDataValidation, StringComparison.InvariantCultureIgnoreCase)
                || vrfCaseStatus.Equals(LegalEntityVrfCaseStatus.RejectedVer1, StringComparison.InvariantCultureIgnoreCase)
                || vrfCaseStatus.Equals(LegalEntityVrfCaseStatus.RejectedVerification, StringComparison.InvariantCultureIgnoreCase)));

Pauljgraham (commented):
one tiny comment

Pauljgraham (approved):

PR #96, opened 2020-10-01
VRF Case status refresh: do not update completed cases

ruslanss:


src/SFA.DAS.EmployerIncentives.Domain/Accounts/LegalEntity.cs
```diff
@@ -6,6 +6,7 @@ namespace SFA.DAS.EmployerIncentives.Domain.Accounts
 {
     public sealed class LegalEntity : Entity<long, LegalEntityModel>
     {
+        private const string VrfCaseStatusCompleted = "Case request complete";
```
David-B-Read, line 9: Can we put the VRF statuses into constants, like this?
https://github.com/SkillsFundingAgency/das-employer-incentives/blob/10425dfb6b044fb20b05b4d97112888968b4c55b/src/SFA.DAS.EmployerIncentives.Domain/Accounts/LegalEntityVrfCaseStatus.cs

PR #103, opened 2020-10-14
Ei 439 store apprenticeship ukprn

David-B-Read:


src/SFA.DAS.EmployerIncentives.Abstractions/DTOs/Commands/IncentiveApplicationApprenticeshipDto.cs
```diff
@@ -12,5 +12,6 @@ public class IncentiveApplicationApprenticeshipDto
         public long Uln { get; set; }
         public DateTime PlannedStartDate { get; set; }
         public ApprenticeshipEmployerType ApprenticeshipEmployerTypeOnApproval { get; set; }
+        public long? ProviderUKPRN { get; set; }
```
MichaelYoung1981, line 15: Doesn't need to be prefixed with provider as that's the P in UKPRN.

David-B-Read: That's updated

Pauljgraham (approved):

MichaelYoung1981 (approved):

PR #106, opened 2020-10-16
EI-440 console app to generate update script for apprenticeship UKPRNs

David-B-Read:


Pauljgraham (commented):
Not sure where this console application would run. I assume DevOps would be expected to run it and then apply the output sql file to EmployerIncentives DB. 

I only ask because they didn't allow us to console applications to do this before, but I'm not surprised  if they have changed their minds

Pauljgraham (approved):

PR #108, opened 2020-10-23
EI-49 domain events and ouw with EI-530  fixed bug on updating applicationApprenticeship record

Pauljgraham:


ruslanss (approved):

src/SFA.DAS.EmployerIncentives.Domain/IncentiveApplications/Apprenticeship.cs
```diff
@@ -18,7 +18,7 @@ public class Apprenticeship : Entity<Guid, ApprenticeshipModel>
         public decimal TotalIncentiveAmount => Model.TotalIncentiveAmount;
         public bool EarningsCalculated => Model.EarningsCalculated;
 
-        internal static Apprenticeship Create(ApprenticeshipModel model)
```
Stephek, line 21: Can we use the internals visible to option to allow the tests to use this?

src/tests/SFA.DAS.EmployerIncentives.Commands.UnitTests/CompleteEarningsCalculation/Handlers/WhenHandlingCompleteEarningsCalculationCommand.cs
```diff
@@ -0,0 +1,75 @@
...
+            var incentiveApplication = _fixture.Create<IncentiveApplication>();
+            var firstApprenticeship = incentiveApplication.Apprenticeships.First();
+            firstApprenticeship.Id = _apprenticeshipIncentiveId;
+            var apprenticeshipModel = _fixture.Build<ApprenticeshipModel>().With(x => x.Id, _apprenticeshipIncentiveId)
+                .Without(x => x.EarningsCalculated).Create();
+            var apprenticeshipModel2 = _fixture.Create<ApprenticeshipModel>();
+
+            var domainIncentiveApplication = _fixture.Create<Domain.IncentiveApplications.IncentiveApplication>();
+
+            domainIncentiveApplication.SetApprenticeships(new List<Apprenticeship> { Apprenticeship.Create(apprenticeshipModel), Apprenticeship.Create(apprenticeshipModel2) });
```
Stephek, line 56: I think it would be better to use AuthoFixture to create the ApplicationModel as required and then use the GetExisting factory method to create the domain object.  Then there is no dependency/knowledge about setting up the domain object in the tests.

PR #118, opened 2020-11-02
Ei 51 earnings resilience check

David-B-Read:


src/SFA.DAS.EmployerIncentives.Data/EarningsResilienceCheck/EarningsResilienceCheckRepository.cs
```diff
@@ -0,0 +1,60 @@
...
+using System.Collections.Generic;
+using System.Threading.Tasks;
+using System;
+using Microsoft.EntityFrameworkCore;
+using SFA.DAS.EmployerIncentives.Domain.IncentiveApplications.Models;
+using SFA.DAS.EmployerIncentives.Data.Map;
+
+namespace SFA.DAS.EmployerIncentives.Data.EarningsResilienceCheck
+{
+    public class EarningsResilienceCheckRepository : IEarningsResilienceCheckRepository
```
MichaelYoung1981, line 13: This repository is working across multiple bounded contexts, it can't.  Shouldn't it just pull all ApplicationApprenticeships where the application has been submitted and the earning calculated flag is false?

src/SFA.DAS.EmployerIncentives.Data/Models/IncentiveApplicationApprenticeship.cs
```diff
@@ -19,5 +22,12 @@ public partial class IncentiveApplicationApprenticeship
         public ApprenticeshipEmployerType ApprenticeshipEmployerTypeOnApproval { get; set; }
         public decimal TotalIncentiveAmount { get; set; }
         public bool EarningsCalculated { get; set; }
+        [Dapper.Contrib.Extensions.Write(false)]
+        public ICollection<ApprenticeshipIncentive> ApprenticeshipIncentives { get; set; }
```
MichaelYoung1981, line 26: Crossing bounded contexts.

src/SFA.DAS.EmployerIncentives.Domain/Services/EarningsResilienceCheckService.cs
```diff
@@ -0,0 +1,7 @@
+﻿
+namespace SFA.DAS.EmployerIncentives.Domain.Services
+{
+    public class EarningsResilienceCheckService : IEarningsResilienceCheckService
```
MichaelYoung1981, line 4: Is this empty class needed?

src/SFA.DAS.EmployerIncentives.Events/EarningsResilienceCheck/EarningsCalculationRequiredHandler.cs
```diff
@@ -0,0 +1,46 @@
...
+                    apprenticeship.ApprenticeshipId,
+                    apprenticeship.FirstName,
+                    apprenticeship.LastName,
+                    apprenticeship.DateOfBirth,
+                    apprenticeship.Uln,
+                    apprenticeship.PlannedStartDate,
+                    apprenticeship.ApprenticeshipEmployerTypeOnApproval
+                );
+
+                var task = _commandPublisher.Publish(command);
```
MichaelYoung1981, line 39: What happens here is the apprenticeship incentive exists but it doesn't have earnings?  A quick look suggests it'll create a duplicate apprenticeship incentive record.

MichaelYoung1981 (changes requested):

src/SFA.DAS.EmployerIncentives.Events/EarningsResilienceCheck/EarningsCalculationRequiredHandler.cs
MichaelYoung1981: We need to do one of two things here...
1) Change CreateApprenticeshipIncentiveCommand so it doesn't create a duplicate record but does publish the Created command.
2) Change this implementation to check for the existence of the ApprenticeshipIncentive and if it doesn't exist publish the CreateApprenticeshipIncentiveCommand but if it does exist then publish a CalculateEarningsCommand

src/SFA.DAS.EmployerIncentives.Commands/ApprenticeshipIncentive/CreateIncentive/CreateApprenticeshipIncentiveCommandHandler.cs
```diff
@@ -0,0 +1,67 @@
...
+                    var calculatePaymentsEvent = new Created
+                    {
+                        AccountId = command.AccountId,
+                        ApprenticeshipId = command.ApprenticeshipId,
+                        ApprenticeshipIncentiveId = existing.Id
+                    };
+                    await _domainEventDispatcher.Send(calculatePaymentsEvent);
+                }
+                return;
+            }
```
MichaelYoung1981, line 48: Domain events needs to be raised from within the domain rather than the application layer.

src/SFA.DAS.EmployerIncentives.Data/EarningsResilienceCheck/EarningsResilienceCheckRepository.cs
```diff
@@ -0,0 +1,50 @@
...
+            earningsCheckApplications.AddRange(applicationsWithoutApprenticeshipIncentives.Select(application => application.Id));
+
+            return await Task.FromResult(earningsCheckApplications);
+        }
+
+        public async Task<IncentiveApplicationModel> GetApplicationDetail(Guid applicationId)
+        {
+            var incentiveApplication = await _dbContext.Applications
+               .Include(x => x.Apprenticeships)
+               .FirstOrDefaultAsync(a => a.Id == applicationId);
```
MichaelYoung1981, line 42: Should be single or default...doubt it makes any difference to the sql generated tho.

David-B-Read: FirstOrDefault will gracefully return null

ruslanss (approved):

src/SFA.DAS.EmployerIncentives.Data/ApprenticeshipIncentives/ApprenticeshipIncentiveDataRepository.cs
```diff
@@ -22,6 +25,29 @@ public async Task Add(ApprenticeshipIncentiveModel apprenticeshipIncentive)
             await _dbContext.SaveChangesAsync();
         }
 
+        public async Task<List<ApprenticeshipIncentiveModel>> FindApprenticeshipIncentivesWithoutPayments()
```
MichaelYoung1981, line 28: Can this be renamed to WithoutPendingPayments?

src/SFA.DAS.EmployerIncentives.Data/IncentiveApplication/IncentiveApplicationDataRepository.cs
```diff
@@ -44,5 +48,52 @@ public async Task Update(IncentiveApplicationModel incentiveApplication)
                 _dbContext.AddRange(model.Apprenticeships);
             }
         }
+
+        public async Task<List<IncentiveApplicationModel>> FindApplicationsWithoutApprenticeshipIncentives()
```
MichaelYoung1981, line 52: Isn't this finding applications without EarningsCalculated?  ApprenticeshipIncentives is a term from another bounded context.

src/SFA.DAS.EmployerIncentives.Events/EarningsResilienceCheck/EarningsCalculationRequiredHandler.cs
MichaelYoung1981: This still looks like a big issue to me?

MichaelYoung1981 (approved):

src/SFA.DAS.EmployerIncentives.Domain/IncentiveApplications/IncentiveApplication.cs
```diff
@@ -79,5 +74,10 @@ private void AddApprenticeship(Apprenticeship apprenticeship)
             _apprenticeships.Add(apprenticeship);
             Model.ApprenticeshipModels.Add(apprenticeship.GetModel());
         }
+
+        public void CalculateEarnings()
```
Stephek, line 78: I don't understand why his method has been put on the domain object?  No state change to the domain object or business rule validation is being done.  Why not just get the command handler to publish the CalculateEarningsCommand ?

src/SFA.DAS.EmployerIncentives.Commands/EarningsResilienceCheck/EarningsResilienceApplicationsCheckCommandHandler.cs
```diff
@@ -0,0 +1,29 @@
...
+        {
+            _applicationDomainRepository = applicationDomainRepository;
+        }
+
+        public async Task Handle(EarningsResilienceApplicationsCheckCommand command, CancellationToken cancellationToken = default)
+        {
+            var applications = await _applicationDomainRepository.FindIncentiveApplicationsWithoutEarningsCalculations();
+            foreach (var application in applications)
+            {
+                application.CalculateEarnings();
```
Stephek, line 23: Why do this?  Why not just publish the CalculateEarningsCommand ?  There is no need to call a method on the domain object to implement the use case unless the domain object is making some business decision. It doesn't make any sense that the application would calculate earnings.  It's got nothing to do with the application.

So maybe just replace with 

var command = new RefreshEarningsCommand(
                application.AccountId,
                application.IncentiveApplicationId,
                application.AccountLegalEntityId);

            return _commandPublisher.Publish(command)

and have a handler in the ApprenticeshipIncentive domain that re-calculates the calculation or creates the incentive if it doesn't exist using the existing command handlers.

Stephek (changes requested):

src/SFA.DAS.EmployerIncentives.Commands/ApprenticeshipIncentive/CreateIncentive/CreateApprenticeshipIncentiveCommandHandler.cs
Stephek: In that case we shouldn't we re-submit the application? So call application.Submit() since the resilience check should re-do what should have already been done.

src/SFA.DAS.EmployerIncentives.Commands/EarningsResilienceCheck/EarningsResilienceApplicationsCheckCommandHandler.cs
Stephek: See above

Stephek (approved):

PR #127, opened 2020-11-10
EI-226 Identify whether an apprenticeship is provider data locked

ruslanss:


src/SFA.DAS.EmployerIncentives.Commands/Services/LearnerMatchApi/LearnerDataExtensionsDataLocks.cs
```diff
@@ -0,0 +1,31 @@
...
+
+            return (from tr in learnerData.Training
+                    where tr.Reference == "ZPROG001"
+                    from pe in tr.PriceEpisodes
+                    from p in pe.Periods
+                    where p.ApprenticeshipId == learner.ApprenticeshipId
+                          && !p.IsPayable
+                          && p.Period == learner.NextPendingPayment.CollectionPeriod
+                          && pe.StartDate <= learner.NextPendingPayment.DueDate
+                          && (pe.EndDate >= learner.NextPendingPayment.DueDate || pe.EndDate == null)
```
MichaelYoung1981, line 27: I really like this lambda but I reckon if some of these conditions were refactored into single line methods it might be more obvious when reading.

src/tests/SFA.DAS.EmployerIncentives.Commands.UnitTests/Services/LearnerService/WhenRefreshCalledAndDataLocksAreChecked.cs
```diff
@@ -0,0 +1,125 @@
...
+
+namespace SFA.DAS.EmployerIncentives.Commands.UnitTests.Services.LearnerService
+{
+    //    Conditions for DataLock:
+    // 1. ZPROG001 reference
+    // 2. PriceEpisode.StartDate <= NexPayment.DueDate => PriceEpisode.EndDate
+    // 3. Period.Period == NexPayment.CollectionPeriod
+    // 4. Period.ApprenticeshipId == ApprenticeshipIncentive.ApprenticeshipId
+    // 5. Period.IsPayable == false
+    public class WhenRefreshCalledAndDataLocksAreChecked
```
MichaelYoung1981, line 20: You probably want a test for each condition in HasProviderDataLocks rather than a single catch all test.

PR #128, opened 2020-11-11
EI 492 - 513 identify if apprentice is in learning

Stephek:


src/SFA.DAS.EmployerIncentives.Commands/Services/LearnerMatchApi/LearnerService.cs
```diff
@@ -40,6 +51,11 @@ public async Task Refresh(Learner learner)
                 learnerSubmissionDto.Training.Any(t => t.Reference == "ZPROG001")
                 );
 
+            submissionData.SetStartDate(learnerSubmissionDto.LearningStartDateForAppenticeship(learner.ApprenticeshipId));
+
+            var firstPendingPayment = incentive.PendingPayments.Where(p => p.PaymentMadeDate == null).OrderBy(p => p.DueDate).FirstOrDefault();
+            submissionData.SetIsInLearning(learnerSubmissionDto.InLearningForAppenticeship(learner.ApprenticeshipId, firstPendingPayment));
+
```
Pauljgraham, line 58: I know I've been off the project for the last week, so I may have missed some conversations and agreements, but I'm a little concerned about the setting of the IsInLearning Flag in this way.

This setting will occur when we refresh the  Learner (I assume that's happening daily or weekly or ....). But what happens when we a change occurs to the learner. Such as the start date or date of birth. This may then need to be recalculated as it would adjust this value.

I originally thought these statuses would be calculated on the fly as we were validating the payments. But as I say I might be a little out of date with the design agreed

Pauljgraham (commented):
A few concerns, when trying to get my head round the way we are working these statuses

src/SFA.DAS.EmployerIncentives.Commands/Services/LearnerMatchApi/LearnerService.cs
Stephek: There are separate stories for change of circumstances.

Pauljgraham (approved):

PR #129, opened 2020-11-11
EI-585 add fields and data to collection calendar

David-B-Read:


src/SFA.DAS.EmployerIncentives.Domain/ValueObjects/CollectionPeriod.cs
```diff
@@ -6,25 +6,34 @@ namespace SFA.DAS.EmployerIncentives.Domain.ValueObjects
...
+            AcademicYear = academicYear;
+            Active = active;
         }
 
         public byte PeriodNumber { get; }
         public byte CalendarMonth { get; }
         public short CalendarYear { get; }
         public DateTime OpenDate { get; }
+        public DateTime CensusDate { get; }
+        public string AcademicYear { get; }
```
Stephek, line 25: Any reason the AcademicYear is a string rather than a short?

Stephek (approved):
Just one question.

src/SFA.DAS.EmployerIncentives.Domain/ValueObjects/CollectionPeriod.cs
David-B-Read: It's not strictly a numeric value, it's a concatenation of two partial year values

Pauljgraham (approved):

PR #130, opened 2020-11-11
Ei 513 unit of work

MichaelYoung1981:


David-B-Read (approved):

src/SFA.DAS.EmployerIncentives.Commands/ServiceCollectionExtensions.cs
```diff
@@ -107,7 +108,8 @@ public static IServiceCollection AddCommandHandlerDecorators(this IServiceCollec
                 .Decorate(typeof(ICommandHandler<>), typeof(CommandHandlerWithDistributedLock<>))
                 .Decorate(typeof(ICommandHandler<>), typeof(CommandHandlerWithRetry<>))
                 .Decorate(typeof(ICommandHandler<>), typeof(CommandHandlerWithValidator<>))
-                .Decorate(typeof(ICommandHandler<>), typeof(CommandHandlerWithLogging<>));
+                .Decorate(typeof(ICommandHandler<>), typeof(CommandHandlerWithLogging<>))
+                .Decorate(typeof(ICommandHandler<>), typeof(CommandHandlerWithUnitOfWork<>));
```
Stephek, line 112: I think this should be the inner decorator.  So the order would be 

CommandHandlerWithLogging -> CommandHandlerWithValidator -> CommandHandlerWithRetry -> CommandHandlerWithDistributedLock -> CommandHandlerWithUnitOfWork -> handler

So needs to be moved to the first in the list.

Stephek (changes requested):
Looks good.  Just one comment.

src/SFA.DAS.EmployerIncentives.Commands/ServiceCollectionExtensions.cs
```diff
@@ -236,18 +238,26 @@ public static IServiceCollection AddLearnerService(this IServiceCollection servi
             return serviceProvider;
         }
 
-        public static IServiceCollection AddEntityFrameworkForEmployerIncentives(this IServiceCollection services)
+        public static IServiceCollection AddEntityFrameworkForEmployerIncentives(this IServiceCollection services, IConfiguration configuration)
```
Stephek, line 241: Is configuration being used?

Stephek (changes requested):

Stephek (approved):

PR #131, opened 2020-11-12
Ei 532 identify pending payments

David-B-Read:


src/SFA.DAS.EmployerIncentives.Data/ApprenticeshipIncentives/Models/PendingPayment.cs
```diff
@@ -18,5 +18,6 @@ public partial class PendingPayment
         public short? PeriodNumber { get; set; }
         public short? PaymentYear { get; set; }
         public long AccountLegalEntityId { get; set; }
+        public short? EarningType { get; set; }
```
MichaelYoung1981, line 21: Can't we use an enum?

MichaelYoung1981: Stored as a string in the db, same way we do with application status.

David-B-Read: We can, the ticket was a little vague on what should actually be stored so had to guess :)

MichaelYoung1981 (approved):

PR #132, opened 2020-11-12
Ei 535 ei 490 ei 226 data locks acceptance tests

ruslanss:


src/tests/SFA.DAS.EmployerIncentives.Functions.PaymentsProcess.AcceptanceTests/Services/TestPaymentsProcessFunctions.cs
```diff
@@ -1,37 +1,37 @@
...
+        private const int OrchestratorRunTimeoutInSeconds = 10;
+
         private bool _isDisposed;
-        private Process _functionHostProcess;
+        private readonly HttpClient _client;
+        private readonly FunctionsController _functionsController;
+        private readonly TestPaymentsProcessFunctionsConfigurator _configurator;
 
-        public TestPaymentsProcessFunctions(TestContext testContext)
+        public TestPaymentsProcessFunctions(string databaseConnectionString, string learnerMatchApiBaseUrl)
```
Stephek, line 24: I think you should still pass in just the TestContext and get the information required from it.  Then you are covered for future changes or if other scenarios need extra information.

PR #135, opened 2020-11-16
Ei 566 still in learning validation check

Pauljgraham:
Branched off EI-492-513-identify-if-apprentice-is-in-learning


src/SFA.DAS.EmployerIncentives.Api.AcceptanceTests/SFA.DAS.EmployerIncentives.Api.AcceptanceTests.csproj
```diff
@@ -5,6 +5,14 @@
     <IsPackable>false</IsPackable>
   </PropertyGroup>
 
+  <ItemGroup>
```
Stephek, line 8: Can you delete these.

src/tests/SFA.DAS.EmployerIncentives.Functions.PaymentsProcess.AcceptanceTests/Features/ValidatePayments.feature
```diff
@@ -8,15 +8,24 @@ Feature: ValidatePayments
...
+	And the apprentice 'is in learning' is <IsInLearning> 
 	When the payment process is run
-	Then failed validation results are recorded
+	Then bank details validation check is <BankDetailsValidationResult>
+	Then apprentice is in learning check is <IsInLearningValidationResult>
 	And no payment records are created
 	And pending payments are not marked as paid
+
+Examples:
+	| BankDetailStatus | IsInLearning | BankDetailsValidationResult | IsInLearningValidationResult |
```
Stephek, line 28: Do we want to do this or just keep every scenario separate?  There will be quite a few more validation states so this would end up with a lot of permutations.  

Then have a separate test for all validations steps passed?

Stephek (changes requested):
Just a comment for discussion.

src/tests/SFA.DAS.EmployerIncentives.Functions.PaymentsProcess.AcceptanceTests/Features/ValidatePayments.feature
Pauljgraham: I was thinking about this, and the tests examples can just allow every validation to pass except one. This will keep the number of permutations down to the minimum. The only permutation I can see is when we have a status of null or false (such as the IsInLearning flag).

src/SFA.DAS.EmployerIncentives.Api.AcceptanceTests/SFA.DAS.EmployerIncentives.Api.AcceptanceTests.csproj
Pauljgraham: This intrigues me. I noticed Dave had added this exclusion to the Functions Acceptance tests and they were working on my machine without the continual deletion of the '**' files. So although I really don't like it, does it cause us a problem leaving it in?

src/tests/SFA.DAS.EmployerIncentives.Functions.PaymentsProcess.AcceptanceTests/Features/ValidatePayments.feature
Stephek: So in that case how about changing it to something like

Given a pending payment
When the "validationstep" validation fails
Then the "validationstep" validation result is recorded
And no payment records are created
And pending payments are not marked as paid

Examples

| validationstep | 
| BankDetailStatus |
| IsInLearning |
| HasLearning |
| ... |

Then each test is distinct and handle he data set up in the code rather than in the specification

Stephek (changes requested):

src/tests/SFA.DAS.EmployerIncentives.Functions.PaymentsProcess.AcceptanceTests/Features/ValidatePayments.feature
Pauljgraham: agreed

PR #141, opened 2020-11-18
API acceptance test performance: use `model` database as template

ruslanss:


David-B-Read (approved):

Pauljgraham (approved):

Pauljgraham (commented):
only concern is we are updating the sql model DB with these tables and therefore if we use this sql instance for anything other than tests, it may cause problems there

ruslanss:
> only concern is we are updating the sql model DB with these tables and therefore if we use this sql instance for anything other than tests, it may cause problems there

Hi Paul, good point. I have spoken to Adam DevOps who advised the container image is reset every time, hence we should be getting a new instance of the localDb on every build. 🤞

PR #145, opened 2020-11-19
EI-223 Assign subnominal code to payment record based on age and levy status

Pauljgraham:
…rtOfCourse


src/SFA.DAS.EmployerIncentives.Domain.UnitTests/ApprenticeshipIncentive/WhenCreatePayment.cs
```diff
@@ -54,8 +57,38 @@ public void Then_the_payment_is_not_created_when_the_pending_payment_is_not_vali
...
+
+            // assert
+            _sut.Payments.Count.Should().Be(1);
+            var actualPayment = _sut.Payments.First();
+            actualPayment.SubnominalCode.Should().Be(expectedSubnominalCode);
+        }
+
         [Test]
-        public void Then_the_payment_is_created()
+        public void Then_the_payment_is_created_with_correct_SubnominalCode()
```
David-B-Read, line 91: The test name implies a check for subnominal code, but this is done in another test

David-B-Read (approved):

PR #157, opened 2020-11-26
Created a new view to expose the business data

DavidBSteele:
This view is so that I can produce a business focus dashboard which exposes the numbers that we need to report to other areas of the business. The current applications dashboard is not suitable for this.

The method for splitting by payment value is a surrogate for recalculating the age. This will give the business the split they need but without adding logic in the view to cal the age. I didn't want to have the logic in multiple places.

Branch name doesn't include the ticket as the ticket was made after the work started

https://skillsfundingagency.atlassian.net/browse/EI-655

### DIFF
--- a/src/SFA.DAS.EmployerIncentives.Database/Views/BusinessDashboardApplications.sql
+++ b/src/SFA.DAS.EmployerIncentives.Database/Views/BusinessDashboardApplications.sql
@@ -1,0 +1,27 @@
+﻿CREATE VIEW [dbo].[BusinessDashboardApplications]
+	AS 
+		select
+			month(PlannedStartDate) as [Planned Start Month],
+			year(PlannedStartDate) as [Planned Start Year],
+			count(distinct(iaa.incentiveapplicationid)) as Applications, 
+			avg(cast(Learners as float)) as [Mean learners per app],
+			count(Learners) as [Num Learners],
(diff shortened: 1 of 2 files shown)


src/SFA.DAS.EmployerIncentives.Database/Views/BusinessDashboardApplications.sql
```diff
@@ -0,0 +1,27 @@
...
+							status
+							FROM [dbo].[IncentiveApplicationApprenticeship] iaa2
+							left join [dbo].[IncentiveApplication] ia2 on ia2.Id=iaa2.IncentiveApplicationId
+							group by [IncentiveApplicationId], status
+						  ) q on  q.IncentiveApplicationId = iaa.IncentiveApplicationId
+		where
+			q.status = 'Submitted'
+		group by 
+			month(PlannedStartDate), year(PlannedStartDate)
+		order by year(PlannedStartDate), month(PlannedStartDate)
```
ruslanss: one cannot order by in views ❌

ruslanss (approved):

src/SFA.DAS.EmployerIncentives.Database/Views/BusinessDashboardApplications.sql
DavidBSteele: Agreed...changed

PR #159, opened 2020-11-27
Ei 610 set active collection calendar period

David-B-Read:


src/SFA.DAS.EmployerIncentives.Api/Controllers/CollectionCalendarCommandController.cs
```diff
@@ -0,0 +1,25 @@
...
+{
+    [Route("collectionCalendar")]
+    [ApiController]
+    public class CollectionCalendarCommandController : ApiCommandControllerBase
+    {
+        public CollectionCalendarCommandController(ICommandDispatcher commandDispatcher) : base(commandDispatcher)
+        {
+        }
+
+        [HttpPost]
```
Stephek, line 17: POST is used to create data.  We are changing an existing record so should be a PATCH or a PUT to the Periods endpoint.  The "IsActive" state should be in the data along with the period number and year since that is what is being changed.  Otherwise we will end up with an RPC style API which creates extra coupling.

Stephek (changes requested):

src/SFA.DAS.EmployerIncentives.Api/Controllers/CollectionCalendarCommandController.cs
David-B-Read: I disagree about having the IsActive state in the data, that would require the calling Azure function to retrieve all the collection periods in order to set all the other periods to inactive and set the new period to active. Happy to change the Http method though.

Stephek: The caller wouldn't have to retrieve all other collection periods.  Because IsActive is set on this one means the others will be set to false as a result since there can only be one that IsActive.  By putting it in the data means that we can easily meet future requirements without breaking or changing the interface every time.

David-B-Read: Updated the API to allow for de-activation of collection periods for the kill switch scenario as discussed.

src/SFA.DAS.EmployerIncentives.Api/Controllers/CollectionCalendarCommandController.cs
```diff
@@ -0,0 +1,27 @@
...
+    [Route("collectionCalendar")]
+    [ApiController]
+    public class CollectionCalendarCommandController : ApiCommandControllerBase
+    {
+        public CollectionCalendarCommandController(ICommandDispatcher commandDispatcher) : base(commandDispatcher)
+        {
+        }
+
+        [HttpPatch]
+        [Route("period/active")]
```
Stephek, line 18: This is just "collectionPeriods"  the operation is a patch and it is an instance that is being updated.  The data contains the change not the endpoint.

src/SFA.DAS.EmployerIncentives.Commands/CollectionCalendar/ActivateCollectionPeriodCommand.cs
```diff
@@ -0,0 +1,18 @@
+﻿using SFA.DAS.EmployerIncentives.Abstractions.Commands;
+
+namespace SFA.DAS.EmployerIncentives.Commands.CollectionPeriod
+{
+    public class ActivateCollectionPeriodCommand : ICommand
```
Stephek, line 5: This should be UpdateCollectionPeriodCommand

src/SFA.DAS.EmployerIncentives.Commands/CollectionCalendar/ActivateCollectionPeriodCommandHandler.cs
```diff
@@ -0,0 +1,26 @@
...
+        public ActivateCollectionPeriodCommandHandler(ICollectionCalendarService collectionCalendarService)
+        {
+            _collectionCalendarService = collectionCalendarService;
+        }
+
+        public async Task Handle(ActivateCollectionPeriodCommand command, CancellationToken cancellationToken = default)
+        {
+            var collectionCalendar = await _collectionCalendarService.Get();
+
+            collectionCalendar.ActivatePeriod(command.CollectionPeriodYear, command.CollectionPeriodNumber, command.Active);
```
Stephek, line 21: Pass in a CollectionPeriod value object.  Note the existing CollectionPeriod  has too many properties on it's constructor.  there is a tech debt item for that.   So for now create a new constructor with just periodNumber , and year.

collectionCalendar.ActivatePeriod could be just collectionCalendar.SetActive(CollectionPeriod  period)

and add a condition to see if command.Activeis true

i.e.

if(command.Active)
{
 collectionCalendar.SetActive(CollectionPeriod  period)
}

// else just ignore since there is no use case for updating anything other than active CollectionPeriods yet.

src/SFA.DAS.EmployerIncentives.Domain/ValueObjects/CollectionCalendar.cs
```diff
@@ -32,6 +32,31 @@ public CollectionPeriod GetPeriod(short collectionYear, byte periodNumber)
...
+                foreach (var collectionPeriod in _collectionPeriods)
+                {
+                    collectionPeriod.SetActive(false);
+                }
+            }
+
+            collectionPeriodToActivate.SetActive(setToActive);
+        }
+
+        public IEnumerable<CollectionPeriod> GetAllPeriods()
```
Stephek, line 55: Should return a readonly collection.

src/SFA.DAS.EmployerIncentives.Commands/CollectionCalendar/ActivateCollectionPeriodCommandHandler.cs
```diff
@@ -0,0 +1,26 @@
+﻿using SFA.DAS.EmployerIncentives.Abstractions.Commands;
+using SFA.DAS.EmployerIncentives.Commands.CollectionPeriod;
+using SFA.DAS.EmployerIncentives.Commands.Services;
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace SFA.DAS.EmployerIncentives.Commands.CollectionCalendar
+{
+    public class ActivateCollectionPeriodCommandHandler : ICommandHandler<ActivateCollectionPeriodCommand>
```
Stephek, line 9: I think this should be UpdateCollectionPeriodCommandHandler.

src/SFA.DAS.EmployerIncentives.Domain/ValueObjects/CollectionCalendar.cs
```diff
@@ -32,6 +32,31 @@ public CollectionPeriod GetPeriod(short collectionYear, byte periodNumber)
                 .Single(d => d.CalendarYear == collectionYear && d.PeriodNumber == periodNumber);
         }
 
+        public void ActivatePeriod(short collectionYear, byte periodNumber, bool setToActive)
```
Stephek, line 35: Could be just SetActive(CollectionPeriod period).

Stephek (changes requested):

Stephek (approved):

PR #163, opened 2020-12-01
EI-552 Send Payments to Business Central Payment API

Pauljgraham:


src/SFA.DAS.EmployerIncentives.Commands/Services/BusinessCentralApi/BusinessCentralFinancePaymentsService.cs
```diff
@@ -0,0 +1,125 @@
...
+
+            if (response.StatusCode == HttpStatusCode.Accepted)
+            {
+                var morePaymentsToSend = payments.Count > paymentsToSend.Count;
+                return new PaymentsSuccessfullySent(paymentsToSend, !morePaymentsToSend);
+            }
+
+            if (response.StatusCode >= HttpStatusCode.InternalServerError)
+            {
+                throw new BusinessCentralApiException($"Business Central API is unavailable and returned an internal code of {response.StatusCode}", response.StatusCode);
```
Stephek, line 43: Can you move the logic to set the message into the BusinessCentralApiException class? i.e. pass the HttpStatus code into the constructor of the exception class and let that determine the message to be set.

src/SFA.DAS.EmployerIncentives.Commands/Services/BusinessCentralApi/BusinessCentralFinancePaymentsService.cs
```diff
@@ -0,0 +1,125 @@
...
+        {
+            _client = client;
+            _apiVersion = apiVersion ?? "2020-10-01";
+            _paymentRequestsLimit = paymentRequestsLimit <= 0 ? 1000 : paymentRequestsLimit;
+        }
+
+        public async Task<PaymentsSuccessfullySent> SendPaymentRequestsForLegalEntity(List<PaymentDto> payments)
+        {
+            var paymentsToSend = payments.Take(_paymentRequestsLimit).ToList();
+            var content = CreateJsonContent(paymentsToSend);
```
Stephek, line 32: Can you not use PostAsJsonAsync ?

src/SFA.DAS.EmployerIncentives.Commands/ServiceCollectionExtensions.cs
```diff
@@ -183,6 +192,31 @@ public static IServiceCollection AddAccountService(this IServiceCollection servi
...
 
+        private static IServiceCollection AddBusinessCentralClient<T>(this IServiceCollection serviceCollection, Func<HttpClient, IServiceProvider, string, int, T> instance) where T : class
+        {
+            serviceCollection.AddTransient(s =>
+            {
+                var settings = s.GetService<IOptions<BusinessCentralApiClient>>().Value;
+
+                var clientBuilder = new HttpClientBuilder()
+                    .WithDefaultHeaders()
+                    .WithApimAuthorisationHeader(settings)
```
Stephek, line 203: Just checking this is calling the APIM?  Or should it be using the Managed Identity Authorisation?

src/SFA.DAS.EmployerIncentives.Commands/Services/BusinessCentralApi/PaymentsSuccessfullySent.cs
```diff
@@ -0,0 +1,16 @@
+﻿using System.Collections.Generic;
+using SFA.DAS.EmployerIncentives.Abstractions.DTOs.Queries.ApprenticeshipIncentives;
+
+namespace SFA.DAS.EmployerIncentives.Commands.Services.BusinessCentralApi
+{
+    public class PaymentsSuccessfullySent
```
Stephek, line 6: Can this be renamed as SendPaymentsResponse since the data contains what was successfully sent?

Stephek (changes requested):

src/SFA.DAS.EmployerIncentives.Commands/ApprenticeshipIncentive/SendPaymentRequests/SendPaymentRequestsCommandHandler.cs
```diff
@@ -0,0 +1,46 @@
...
+            if (payments.Any() == false)
+            {
+                return;
+            }
+
+            var sent = await _businessCentralFinancePaymentsService.SendPaymentRequestsForLegalEntity(payments);
+
+            await _accountRepository.UpdatePaidDateForPaymentIds(sent.PaymentsSent.Select(s => s.PaymentId).ToList(), command.AccountLegalEntityId, command.PaidDate);
+
+            if (!sent.AllPaymentsSent)
```
Stephek, line 40: Shouldn't this be handled in the IBusinessCentralFinancePaymentsService after all it's an implementation detail of that service that updates need to be paged.  So this handler should just call 

await _businessCentralFinancePaymentsService.SendPaymentRequestsForLegalEntity(payments); 
then 
 await _accountRepository.UpdatePaidDateForPaymentIds(sent.PaymentsSent.Select(s => s.PaymentId).ToList(), command.AccountLegalEntityId, command.PaidDate);

src/SFA.DAS.EmployerIncentives.Commands/Services/BusinessCentralApi/BusinessCentralFinancePaymentsService.cs
```diff
@@ -0,0 +1,125 @@
...
+            var response = await _client.PostAsync($"payments/requests?api-version={_apiVersion}", content);
```
Stephek, line 33: Is this call idempotent?

Stephek (changes requested):

src/SFA.DAS.EmployerIncentives.Commands/ServiceCollectionExtensions.cs
Pauljgraham: No it's an APIM call. But not to our APIM (its the BC one)

src/SFA.DAS.EmployerIncentives.Commands/ApprenticeshipIncentive/SendPaymentRequests/SendPaymentRequestsCommandHandler.cs
Pauljgraham: The original intention was that this command handler would be called via a ServiceBus Message (which I think is still the longer term aim). And if the BC Api could not handle the number of payment requests, we would simply send the same SB message from within the handler. 

With the SB pattern, it meant we wouldn't spend a long time within the command handler, and we could be certain that the messages will eventually be sent to BC. It also meant that if BC was responding slowly or was down, our durable functions wouldn't start retrying the activity and sending even more requests to BC. 

However, mike asked that we didn't implement that yet. Hence this was the compromise pattern I put in place as it closely matches the original intent. I believe a subsequent story is on the backlog which might implement the SB message

src/SFA.DAS.EmployerIncentives.Commands/Services/BusinessCentralApi/BusinessCentralFinancePaymentsService.cs
Pauljgraham: I believe it will accept the same payment requests (as long as they are formatted correctly) and return a 202. However, I understand if the payment request (for a given id) has already been processed it will simply skip over the new details

src/SFA.DAS.EmployerIncentives.Commands/Services/BusinessCentralApi/BusinessCentralFinancePaymentsService.cs
Pauljgraham: Good idea, done

src/SFA.DAS.EmployerIncentives.Commands/ApprenticeshipIncentive/SendPaymentRequests/SendPaymentRequestsCommandHandler.cs
Stephek: In that case can the filter to select the payments be done in the handler?  So move this line

 var paymentsToSend = payments.Take(_paymentRequestsLimit).ToList();

It just seems that the handler and the service have to work together to do the paged updates rather than just one being in control.  Would have to pass in the Options config into the handler to determine the limit

Stephek (changes requested):

src/SFA.DAS.EmployerIncentives.Commands/Services/BusinessCentralApi/BusinessCentralFinancePaymentsService.cs
Pauljgraham: Tried it using PostAsJsonAsync and the problem was the camel case format wasn't getting created and teh BC api was rejecting it

src/SFA.DAS.EmployerIncentives.Commands/Services/BusinessCentralApi/PaymentsSuccessfullySent.cs
Pauljgraham: done

src/SFA.DAS.EmployerIncentives.Commands/Services/BusinessCentralApi/IBusinessCentralFinancePaymentsService.cs
```diff
@@ -0,0 +1,11 @@
+﻿using System.Collections.Generic;
+using System.Threading.Tasks;
+using SFA.DAS.EmployerIncentives.Abstractions.DTOs.Queries.ApprenticeshipIncentives;
+
+namespace SFA.DAS.EmployerIncentives.Commands.Services.BusinessCentralApi
+{
+    public interface IBusinessCentralFinancePaymentsService
+    {
+        public Task<SendPaymentsResponse> SendPaymentRequestsForLegalEntity(List<PaymentDto> payments);
```
Stephek, line 9: Can this be just SendPayments ?  There is nothing in the parameter list that has anything to do with LegalEntities. The legal entity filtering has already happened elsewhere so it doesn't make sense to include it on the method name.

Stephek (changes requested):

src/SFA.DAS.EmployerIncentives.Commands/ApprenticeshipIncentive/SendPaymentRequests/SendPaymentRequestsCommandHandler.cs
Pauljgraham: good point, done

src/SFA.DAS.EmployerIncentives.Commands/Services/BusinessCentralApi/IBusinessCentralFinancePaymentsService.cs
Pauljgraham: done

Stephek (approved):

Stephek (approved):

PR #164, opened 2020-12-01
EI-661-ensure-validation-is-complete-before-create-payment

Stephek:


Pauljgraham (approved):
Did the same thing in 552

PR #165, opened 2020-12-03
Ei 674 add polly retry on transient http errors 

Pauljgraham:


Stephek (commented):
Could be implemented more simply fitting in with the existing patterns


1. create this handler

public sealed class TransientRetryHandler : PolicyHttpMessageHandler
        {
            public TransientRetryHandler(Func<PolicyBuilder<HttpResponseMessage>, IAsyncPolicy<HttpResponseMessage>> policy)
                :base (policy(HttpPolicyExtensions.HandleTransientHttpError()))
            {
            }
        }

2. add to the client like this

var clientBuilder = new HttpClientBuilder()
                    .WithDefaultHeaders()
                    .WithHandler(new TransientRetryHandler(p => p.RetryAsync(3)))

PR #167, opened 2020-12-04
EI-680 Exclude withdrawn incentives from earnings calcs

ruslanss:


src/SFA.DAS.EmployerIncentives.Commands/ApprenticeshipIncentive/CalculateEarnings/CalculateEarningsCommandHandler.cs
```diff
@@ -26,6 +26,7 @@ public class CalculateEarningsCommandHandler : ICommandHandler<CalculateEarnings
         public async Task Handle(CalculateEarningsCommand command, CancellationToken cancellationToken = default)
         {
             var incentive = await _domainRepository.Find(command.ApprenticeshipIncentiveId);
+            if (incentive.Withdrawn) return;
```
Stephek, line 29: This "business logic" should go into the domain object .  To avoid calling the other two services unnecessarily change  the CalculateEarnings method to accept the IIncentivePaymentProfilesService  and ICollectionCalendarService parameters instead of the data.  Then it's the double dispatch pattern and all the business logic can be encapsulated in the domain object.

src/SFA.DAS.EmployerIncentives.Domain/ApprenticeshipIncentives/ApprenticeshipIncentive.cs
```diff
@@ -24,6 +24,8 @@ public sealed class ApprenticeshipIncentive : AggregateRoot<Guid, Apprenticeship
         public PendingPayment NextDuePayment => GetNextDuePayment();
         public IReadOnlyCollection<Payment> Payments => Model.PaymentModels.Map().ToList().AsReadOnly();
 
+        public bool Withdrawn => Model.Withdrawn;
```
Stephek, line 27: Can keep this private if the above change is done.  This avoids coupling on this method.

src/SFA.DAS.EmployerIncentives.Api.AcceptanceTests/Features/ApprenticeshipIncentiveCreated.feature
```diff
@@ -24,4 +24,9 @@ Scenario: Apprenticeship incentive earnings are calculated
 Scenario: Incentive application updated after earnings calculated
 	Given an apprenticeship incentive earnings have been calculated
 	When the earnings calculation against the apprenticeship incentive completes
-	Then the incentive application is updated to record that the earnings have been calculated
\ No newline at end of file
+	Then the incentive application is updated to record that the earnings have been calculated
+
+Scenario: Apprenticeship incentive is withdrawn
+	Given an apprenticeship incentive exists
+	When the apprenticeship incentive is withdrawn
```
Stephek, line 31: The When is what happens so should be the same as the others.  So "When the earnings calculation against the apprenticeship completes".  The Given is the starting point so is that the apprenticeship is withdrawn .

Stephek (changes requested):
A couple of points.

src/SFA.DAS.EmployerIncentives.Api.AcceptanceTests/Features/ApprenticeshipIncentiveCreated.feature
Stephek, line 31: It's the application that is withdrawn not the incentive.

src/SFA.DAS.EmployerIncentives.Commands/ApprenticeshipIncentive/CalculateEarnings/CalculateEarningsCommandHandler.cs
Stephek, line 29: It's the application that is withdrawn not the incentive.  That get's deleted when the application is withdrawn.

src/SFA.DAS.EmployerIncentives.Data/ApprenticeshipIncentives/Models/ApprenticeshipIncentive.cs
```diff
@@ -28,6 +28,8 @@ public partial class ApprenticeshipIncentive
         [Dapper.Contrib.Extensions.Write(false)]
         public ICollection<Payment> Payments { get; set; }
 
+        public bool Withdrawn { get; set; }
```
Stephek, line 31: Should be on the application not the incentive.

src/SFA.DAS.EmployerIncentives.Database/Tables/ApprenticeshipIncentive.sql
```diff
@@ -11,7 +11,8 @@
 	[PlannedStartDate] DATETIME2 NOT NULL,
 	[IncentiveApplicationApprenticeshipId] UNIQUEIDENTIFIER NOT NULL,
 	[AccountLegalEntityId] BIGINT NULL, 
-    [UKPRN] BIGINT NULL
+    [UKPRN] BIGINT NULL, 
+    [Withdrawn] BIT NOT NULL DEFAULT 0
```
Stephek, line 15: Wrong table.

Stephek (changes requested):

src/SFA.DAS.EmployerIncentives.Database/Tables/ApprenticeshipIncentive.sql
David-B-Read: Withdrawals are at an apprenticeship level, so the Withdrawn flag should be on the IncentiveApplicationApprenticeship table

src/SFA.DAS.EmployerIncentives.Commands/ApprenticeshipIncentive/CalculateEarnings/CalculateEarningsCommandHandler.cs
ruslanss: agreed

PR #173, opened 2020-12-10
EI-729 obfuscate uln when sending to BC in certain environments 

Pauljgraham:


src/SFA.DAS.EmployerIncentives.Infrastructure/Configuration/BusinessCentralApiClient.cs
```diff
@@ -9,6 +9,7 @@ public class BusinessCentralApiClient : IApimClientConfiguration
         public string SubscriptionKey { get; set; }
         public string ApiVersion { get; set; }
         public int PaymentRequestsLimit { get; set; }
+        public bool ObfuscateSensitiveData { get; set; }
```
Stephek, line 12: Should it be more specific?  i.e. ObfuscateULN ?

Pauljgraham: I thought about that, but what if we want to hide other data like a name or something else. Though  more general name was better

David-B-Read: I think keep the name generic, in case we need to obfuscate other data in the future

David-B-Read (approved):

PR #188, opened 2020-12-15
Ei 677 compliance withdrawals branched from 676

Stephek:


src/SFA.DAS.EmployerIncentives.Api/Controllers/WithdrawalCommandController.cs
```diff
@@ -11,15 +14,29 @@ namespace SFA.DAS.EmployerIncentives.Api.Controllers
...
         {
+            var incentiveModel = await _queryRepository.Get((a => 
+                a.AccountLegalEntityId == request.AccountLegalEntityId &&
+                a.ULN == request.ULN), includePayments:true);
+            
+            if(incentiveModel != null && incentiveModel.Payments.Any())
+            {
+                return BadRequest(new { Error = "Cannot withdraw an application that has been submitted and has received payments" });
+            }
+
```
David-B-Read, line 39: It doesn't feel right to be calling the repository directly from the controller. Would it be better that the handler performs this check and bubbles an exception up to the controller, which could then be returned as a BadRequest response?

Stephek: It's in a different domain so the handler isn't the correct place.  If the repository is considered as a read only model then I would say it's OK to be called since the client can't change the data.  It's quite common to have a read only repository return value objects as part of CQRS.  e.g. https://stackoverflow.com/questions/47879119/ddd-read-only-repositories-returning-value-objects

David-B-Read: Makes sense.

David-B-Read (approved):

src/SFA.DAS.EmployerIncentives.Api/Controllers/WithdrawalCommandController.cs
```diff
@@ -31,10 +44,35 @@ public async Task<IActionResult> WithdrawalIncentiveApplication([FromBody] Withd
...
+            }
             else
             {
                 return BadRequest();
             }
         }       
+
+        private async Task<bool> HasPayments(long accountLegalEntityId, long uln)
+        {
+            var incentiveModel = await _queryRepository.Get((a =>
```
ruslanss, line 66: a job for a query-handler (which can be used with Value Types) ?

ruslanss (changes requested):

src/SFA.DAS.EmployerIncentives.Api/Controllers/WithdrawalCommandController.cs
ruslanss: can't a QueryHandler be used in a CommandController?

src/SFA.DAS.EmployerIncentives.Api/Controllers/WithdrawalCommandController.cs
Stephek: Maybe but Is that adding extra layers unnecessarily? The IQueryRepository<T> as implemented by  IApprenticeshipIncentiveQueryRepository supports read only querying so what value would a handler add?

Stephek: I can update to fit in with elsewhere if required.

Stephek: Updated

ruslanss (approved):

PR #189, opened 2020-12-16
Ei 723 off ei-678 resume payments

Pauljgraham:
Compared this against a base branch of EI-676 as I refactored some of the pause functionality (and endpoint route), and comparing to this branch makes it easier to see combined functionality of pause/resume payments


Pauljgraham:
branch name is wrong

PR #199, opened 2020-12-21
Ei 733 get vendor id on vrf status update

David-B-Read:


src/SFA.DAS.EmployerIncentives.Api/Controllers/AccountQueryController.cs
```diff
@@ -57,5 +60,19 @@ public async Task<IActionResult> GetApplications(long accountId)
 
             return NotFound();
         }
+
+        [HttpGet("/legalentities/{hashedLegalEntityId}/employervendorid")]
```
Stephek, line 64: Wouldn't it be better and more RESTful to return a GetLegalEntityResponse?  Otherwise we are tying the API to a specific narrow use case.  If the use case changes then it would be a breaking change to the interface.

Could be "/legalentities/{hashedLegalEntityId}?employervendorid=123 to return the LegalEntity object for the vendorId.

Stephek (changes requested):

src/SFA.DAS.EmployerIncentives.Api/Controllers/AccountQueryController.cs
```diff
@@ -57,5 +59,19 @@ public async Task<IActionResult> GetApplications(long accountId)
 
             return NotFound();
         }
+
+        [HttpGet("/legalentities/{hashedLegalEntityId}")]
+        public async Task<IActionResult> GetEmployerVendorId(string hashedLegalEntityId)
```
ruslanss, line 64: s/b GetLegalEntity?

ruslanss (changes requested):

src/SFA.DAS.EmployerIncentives.Api/Controllers/AccountQueryController.cs
ruslanss: can be combined with existing GetLegalEntity in AccountQueryController? i.e. return LegalEntity based on the query string.

David-B-Read: Can't use that for the VRF status update function, only the hashed legal entity id is available to search on. The other endpoint required numeric account and account legal entity ids

ruslanss (approved):

src/SFA.DAS.EmployerIncentives.Api/Controllers/AccountQueryController.cs
Stephek: That's why it should be on the querystring (i.e. a search filter on the collection) rather than the path (which would be the Id of an individual resource)

ruslanss: that's what i was thinking but then it probably won't look great with the if else or switch. perhaps we could pass in the query string to the query handler (somehow) ?

David-B-Read: Updated url

Stephek (approved):

MichaelYoung1981:
This could break our month end bank account validation as all we check for is the existence of the vendor id so if it's populated prior to the case being completed we could end up paying when we shouldn't.

PR #211, opened 2021-01-07
Ei 258 accounts api tech debt

David-B-Read:


src/SFA.DAS.EmployerIncentives.Api/Controllers/AccountCommandController.cs
```diff
@@ -52,5 +53,12 @@ public async Task<IActionResult> AddEmployerVendorId([FromRoute] string hashedLe
             await SendCommandAsync(new AddEmployerVendorIdForLegalEntityCommand(hashedLegalEntityId, request.EmployerVendorId));
             return NoContent();
         }
+
+        [HttpPatch("/legalentities/refresh")]
```
Stephek, line 57: This already exists as a Job resource so not needed as an RPC style interface here.

Stephek (changes requested):

src/SFA.DAS.EmployerIncentives.Api/Controllers/AccountCommandController.cs
David-B-Read: The Job command endpoint uses a Dictionary<string,string> which isn't suitable for what's needed - the outer API now passes in the set of ALEs fetched from the accounts API and the paging info.

Stephek: It uses a Dictionary<string,object> so any json object could be passed to it.

Stephek: Alternatively you could create a "LegalEntitiesRefreshJobs/ endpoint and POST a LegalEntitiesRefreshJob resource to that.  As the technical requirement is to produce a REST API the URL should point to a resource (which would be a noun not a verb).   Refresh isn't a resource of a LegalEntity.

Stephek (approved):

PR #223, opened 2021-01-18
EI-780 display estimated payment month/year

ruslanss:


src/SFA.DAS.EmployerIncentives.Api.AcceptanceTests/DataAccess.cs
```diff
@@ -52,25 +53,10 @@ public async Task SetupAccount(Account account)
...
-                    application.Id,
-                    application.AccountId,
-                    application.AccountLegalEntityId,
-                    application.DateCreated,
-                    Status = application.Status.ToString(),
-                    application.DateSubmitted,
-                    application.SubmittedByEmail,
-                    application.SubmittedByName
-                });
+            return await dbConnection.InsertWithEnumAsString(entity);
```
Stephek, line 59: Dapper has custom TypeMappers e.g.

```
   public class EarningsTypeHandler : SqlMapper.TypeHandler<EarningType>
    {        
        public override EarningType Parse(object value)
        {
            Enum.TryParse(value.ToString(), out EarningType earningType);

            return earningType;
        }

        public override void SetValue(IDbDataParameter parameter, EarningType value)
        {
            parameter.Value = value.ToString();
        }
    }
```

    

   and register with
    
    SqlMapper.AddTypeHandler(typeof(EarningType), new EarningsTypeHandler());
    
    Note doesn't work with Dapper Contrib but will work with SqlMapper.

Stephek (changes requested):

Stephek (approved):

src/SFA.DAS.EmployerIncentives.Api.AcceptanceTests/Steps/ApplicationsForAccountRequestedSteps.cs
```diff
@@ -61,12 +58,13 @@ public void ThenTheApprenticeshipApplicationsAreReturned()
             var apprenticeshipApplication = _apiResponse.ApprenticeApplications.First();
 
             apprenticeshipApplication.AccountId.Should().Be(_account.Id);
-            apprenticeshipApplication.ApplicationId.Should().Be(_application.Id);
-            apprenticeshipApplication.FirstName.Should().Be(_apprenticeship.FirstName);
-            apprenticeshipApplication.LastName.Should().Be(_apprenticeship.LastName);
+            //apprenticeshipApplication.ApplicationId.Should().Be(_application.Id); //TODO: This needs resolving as the bank details needs the original application id
```
Stephek, line 61: TODO :

Stephek (changes requested):

src/SFA.DAS.EmployerIncentives.Api.AcceptanceTests/DataAccess.cs
ruslanss: i've added a generic method which reads the the property attributes to work out whether it needs to insert enum as string or not. please take a look.

Stephek: Nice one.

Can you change the new method to match the InsertAsync  e.g.

```
public static Task<int> InsertAsync<T>(this IDbConnection connection, T entityToInsert, bool enumToString = false, IDbTransaction transaction = null, int? commandTimeout = null, ISqlAdapter sqlAdapter = null) where T : class
        {
            if (enumToString)
            {
                ...
            }
            else
            {
                return Dapper.Contrib.Extensions.SqlMapperExtensions.InsertAsync(connection, entityToInsert, transaction, commandTimeout, sqlAdapter);
            }
        }

```

ruslanss: good idea!

PR #228, opened 2021-01-21
Ei 797 ei 759 learner match check

David-B-Read:


src/SFA.DAS.EmployerIncentives.Data/ApprenticeApplicationDataRepository.cs
```diff
@@ -1,50 +1,73 @@
...
+                              FirstPaymentStatus = data.firstPayment == default ? null : new PaymentStatusDto
+                              {
+                                  PaymentDate = data.firstPayment.DueDate.AddMonths(1),
+                                  LearnerMatchNotFound = (!data.learner.LearningFound.HasValue || !data.learner.LearningFound.Value),
+                                  PaymentAmount = data.firstPayment.Amount
+                              },
+                              SecondPaymentStatus = data.secondPayment == default ? null : new PaymentStatusDto
+                              {
+                                  PaymentDate = data.secondPayment.DueDate.AddMonths(1),
+                                  LearnerMatchNotFound = (!data.learner.LearningFound.HasValue || !data.learner.LearningFound.Value),
```
ruslanss, line 54: wouldn't this fail if data.learner is null?

David-B-Read: No, the DefaultIfEmpty takes care of that.

ruslanss (approved):

PR #229, opened 2021-01-21
Ei 796 agreement version shutter

MichaelYoung1981:


src/SFA.DAS.EmployerIncentives.Domain/ValueObjects/Incentive.cs
```diff
@@ -32,6 +38,17 @@ public Incentive(DateTime dateOfBirth, DateTime startDate, IEnumerable<Incentive
             _payments = GeneratePayments();
         }
 
+        public bool HasSignedRequiredAgreementVersion(int signedagreementVersion)
```
Stephek, line 41: I find this name a bit confusing for a value object.  Does it make sense to call the method CanSignAgreementVersion(int version)?

src/SFA.DAS.EmployerIncentives.Queries/NewApprenticeIncentive/GetApplication/GetApplicationResponse.cs
```diff
@@ -6,9 +6,12 @@ public class GetApplicationResponse
     {
         public IncentiveApplicationDto Application { get; }
 
-        public GetApplicationResponse(IncentiveApplicationDto application)
+        public bool NewAgreementRequired { get; }
```
Stephek, line 9: Shouldn't this go on the IncentiveApplicationDto rather than the GetApplicationResponse?

Stephek (approved):

PR #232, opened 2021-01-21
EI-779 add apprentice not in learning check

David-B-Read:


src/SFA.DAS.EmployerIncentives.Data/ApprenticeApplicationDataRepository.cs
```diff
@@ -47,7 +47,8 @@ public async Task<List<ApprenticeApplicationDto>> GetList(long accountId, long a
                                   PaymentDate = data.firstPayment.DueDate.AddMonths(1),
                                   LearnerMatchNotFound = (!data.learner.LearningFound.HasValue || !data.learner.LearningFound.Value),
                                   PaymentAmount = data.firstPayment.Amount,
-                                  HasDataLock = (data.learner.HasDataLock.HasValue && data.learner.HasDataLock.Value)
+                                  HasDataLock = (data.learner.HasDataLock.HasValue && data.learner.HasDataLock.Value),
+                                  ApprenticeNotInLearning = (data.learner.InLearning.HasValue && !data.learner.InLearning.Value)
```
Stephek, line 51: Do you think it would be better to have this property as just "InLearning".  It against an apprentice application so we don;t need the apprentice bit and if the caller is interested in it being not in learning then they would look for false in the InLearning property.  If they wanted to know if the apprentice was in learning then they'd have to apply the double negative  !ApprenticeNotInLearning

Stephek (changes requested):

src/SFA.DAS.EmployerIncentives.Data/ApprenticeApplicationDataRepository.cs
David-B-Read: Could do. If we're doing that, would make sense to invert the logic on the LearnerMatchNotFound property, for consistency.

Stephek: Yes good idea.

PR #240, opened 2021-01-26
EI 806 813 co c create clawback

Stephek:


David-B-Read (approved):

MichaelYoung1981 (commented):
Couple of things...
1) Doesn't it make sense to have the Clawbacks sitting off the ApprenticeshipIncentive domain object (in the same way as Payments, they're basically the same thing)?
2) The domain events aren't guaranteed to be handled before the orchestrator completes so we could end up running month end before clawbacks have been created (unless we start monitoring queues)

MichaelYoung1981 (approved):

PR #242, opened 2021-01-28
EI-833 Clawbacks - validate unsent clawbacks

ruslanss:


src/SFA.DAS.EmployerIncentives.Domain/ApprenticeshipIncentives/ApprenticeshipIncentive.cs
```diff
@@ -197,11 +197,22 @@ public void CreatePayment(Guid pendingPaymentId, short collectionYear, byte coll
...
+            var existingPayment = Model.PaymentModels.SingleOrDefault(x => x.PendingPaymentId == pendingPaymentId);
+            if (existingPayment != null)
+            {
+                existingPayment.CalculatedDate = paymentDate;
+                existingPayment.PaymentPeriod = collectionPeriod;
+                existingPayment.PaymentYear = collectionYear;
+                existingPayment.SubnominalCode = DetermineSubnominalCode();
+                existingPayment.Amount = pendingPayment.Amount;
+            }
+            else
```
MichaelYoung1981, line 211: When would this ever happen?

MichaelYoung1981 (approved):

src/SFA.DAS.EmployerIncentives.Domain/ApprenticeshipIncentives/ApprenticeshipIncentive.cs
Stephek: When create payment is called the existing code removed the existing payment and added a new one.  So this just updates the existing one.  Whether it can happen I don't know but it's existing functionality and just a reactor to fix a test issue (it's and certainly defensive coding, since this code block shouldn't know how it's being used).  

Are you saying if there is an existing Payment for the Pending Payment Id that the method should just return regardless of the passed in CollectionYear and CollectionPeriod which could be different?

ruslanss: discussed with Michael, this defence can be left in .

PR #248, opened 2021-02-03
EI-864 Enable structured logging

ruslanss:


src/SFA.DAS.EmployerIncentives.Api/NLogConfiguration.cs
```diff
@@ -0,0 +1,75 @@
...
+using NLog.Targets;
+using SFA.DAS.NLog.Targets.Redis.DotNetCore;
+using System;
+using System.IO;
+
+namespace SFA.DAS.EmployerIncentives.Api
+{
+    public class NLogConfiguration
+    {
+        public void ConfigureNLog()
```
Stephek, line 14: Devops prefer us to use nlog.config,  Please update to use that method for configuring nlog.

Stephek (changes requested):

Stephek (approved):

src/SFA.DAS.EmployerIncentives.Api/NLogConfiguration.cs
ruslanss: confirmed they no longer do :)

PR #252, opened 2021-02-09
EI 512 store start date CoC

Stephek:


src/SFA.DAS.EmployerIncentives.Commands/ApprenticeshipIncentive/RefreshLearner/RefreshLearnerCommandHandler.cs
```diff
@@ -63,7 +63,7 @@ public async Task Handle(RefreshLearnerCommand command, CancellationToken cancel
                 submissionData.SetRawJson(learnerData.RawJson);
             }
 
-            if (!submissionData.HasChangeOfCircumstances(learner.SubmissionData))
+            if (submissionData.HasChangeOfCircumstances(learner.SubmissionData))
```
ruslanss, line 66: was this a bug?

Stephek: It was the wrong way round.  Was returning false when the data had a change of circumstance rather than true so was being used the wrong way round too.

ruslanss (approved):

PR #264, opened 2021-03-05
EI 922 - EI 512 with EI 867 learning stopped change of circumstance

Stephek:


src/SFA.DAS.EmployerIncentives.Database/Scripts/Update.ApprenticeshipIncentives.sql
```diff
@@ -3,3 +3,16 @@
 FROM incentives.ApprenticeshipIncentive AI
 INNER JOIN IncentiveApplicationApprenticeship IAA ON IAA.Id = AI.IncentiveApplicationApprenticeshipId
 WHERE AI.UKPRN IS NULL
+GO
+UPDATE i SET
```
David-B-Read, line 7: Would the DACPAC deploy support converting PausePayments into a computed column, to save having to populate the same data in two fields?

David-B-Read (approved):

PR #268, opened 2021-03-10
EI-963 add fields to legal entity to support updating VRF status

David-B-Read:


src/SFA.DAS.EmployerIncentives.Domain/ApprenticeshipIncentives/ValueTypes/LegalEntity.cs
```diff
@@ -24,6 +29,7 @@ protected override IEnumerable<object> GetAtomicValues()
             yield return LegalEntityId;
             yield return LegalEntityName;
             yield return VrfVendorId;
+            yield return HashedLegalEntityId;
```
Stephek, line 32: VrfCaseStatus ?

Stephek (changes requested):

Stephek (approved):

PR #276, opened 2021-03-19
Ei 977 suspend learner match during payment run

MichaelYoung1981:


src/tests/SFA.DAS.EmployerIncentives.Queries.UnitTests/CollectionCalendar/WhenHandlingCollectionPeriodInProgressQuery.cs
```diff
@@ -0,0 +1,51 @@
...
+                new CollectionPeriod(3, 2021)
+            };
+
+            var collectionCalendar = new Domain.ValueObjects.CollectionCalendar(calendarPeriods);
+            _service.Setup(x => x.Get()).ReturnsAsync(collectionCalendar);
+        }
+
+        [TestCase(true)]
+        [TestCase(false)]
+        public async Task ThenActiveCollectionPeriodInProgressStatusIsReturned(bool isActive)
```
David-B-Read, line 42: Param name should be isInProgress?

David-B-Read (approved):

src/SFA.DAS.EmployerIncentives.Functions.PaymentsProcess/Orchestrators/LearnerMatchingOrchestrator.cs
```diff
@@ -23,6 +23,10 @@ public async Task RunOrchestrator([OrchestrationTrigger] IDurableOrchestrationCo
             if(!context.IsReplaying)
                 _logger.LogInformation("LearnerMatchOrchestrator Started");
 
+            var paymentRunInProcess = await context.CallActivityAsync<bool>(nameof(CollectionPeriodInProgress), null);
+            if (paymentRunInProcess)
```
Stephek, line 27: Can we log that the payment has been suspended?

src/SFA.DAS.EmployerIncentives.Functions.PaymentsProcess/Activities/CollectionPeriodInProgress.cs
```diff
@@ -0,0 +1,30 @@
...
+        {
+            _queryDispatcher = queryDispatcher;
+            _logger = logger;
+        }
+
+        [FunctionName(nameof(CollectionPeriodInProgress))]
+        public async Task<bool> Get([ActivityTrigger] object input)
+        {
+            _logger.LogInformation("Getting Collection Period In Progress");
+            var isInProgress = (await _queryDispatcher.Send<CollectionPeriodInProgressRequest, CollectionPeriodInProgressResponse>(new CollectionPeriodInProgressRequest())).IsInProgress;
```
Stephek, line 25: Why put the business functionality in the request?  Would it not be better to just use the exiting GetActiveCollectionPeriodQueryHandler (via GetActiveCollectionPeriod ativity) and add the InProgress property in the response?  Then this very specific query is not required.

src/SFA.DAS.EmployerIncentives.Domain/ValueObjects/CollectionPeriod.cs
```diff
@@ -30,12 +31,18 @@ public CollectionPeriod(byte periodNumber, short academicYear)
         public DateTime CensusDate { get; }
         public short AcademicYear { get; }
         public bool Active { get; private set; }
+        public bool PeriodEndInProgress { get; private set; }
```
Stephek, line 34: The stories refer to Month End Processing.  Shouldn't this be named similarly?  e.g. MonthEndProcessingInProgress ?

src/SFA.DAS.EmployerIncentives.Domain/ValueObjects/CollectionPeriod.cs
```diff
@@ -30,12 +31,18 @@ public CollectionPeriod(byte periodNumber, short academicYear)
...
 
         public void SetActive(bool active)
         {
             Active = active;
         }
 
+        public void SetPeriodEndInProgress(bool periodEndInProgress)
```
Stephek, line 41: This method should be on the CollectionCalendar object since is should set to null any other periods that have the in progress set to true.  I assume only one can be running at a time.

src/SFA.DAS.EmployerIncentives.Domain/ValueObjects/CollectionPeriod.cs
```diff
@@ -6,7 +6,7 @@ namespace SFA.DAS.EmployerIncentives.Domain.ValueObjects
 {
     public class CollectionPeriod : ValueObject
     {
-        public CollectionPeriod(byte periodNumber, byte calendarMonth, short calendarYear, DateTime openDate, DateTime censusDate, short academicYear, bool active)
+        public CollectionPeriod(byte periodNumber, byte calendarMonth, short calendarYear, DateTime openDate, DateTime censusDate, short academicYear, bool active, bool periodEndInProgress)
```
Stephek, line 9: I don't think PeriodEndInProgress should be on the constructor since you shouldn't be able to create one in progress.  Only the domain logic should be able to set to in progress or the data mapper when hydrating.

Stephek (changes requested):

src/SFA.DAS.EmployerIncentives.Domain/ValueObjects/CollectionPeriod.cs
MichaelYoung1981: The mapper uses this method when hydrating.

src/SFA.DAS.EmployerIncentives.Domain/ValueObjects/CollectionPeriod.cs
MichaelYoung1981: I'm tempted to just call it IsInProgress?

Stephek (approved):

PR #280, opened 2021-03-26
Ei 980 suspend earnings calc during payment run

MichaelYoung1981:


David-B-Read (approved):

src/SFA.DAS.EmployerIncentives.Abstractions/Commands/IScheduledCommandPublisher.cs
```diff
@@ -0,0 +1,11 @@
+﻿using System;
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace SFA.DAS.EmployerIncentives.Abstractions.Commands
+{
+    public interface IScheduledCommandPublisher
```
Stephek, line 7: Do we need a separate interface?  Can't the method go on the ICommandPublisher and the delay is implied by the method arguments,

Stephek (changes requested):

src/SFA.DAS.EmployerIncentives.Abstractions/Commands/IScheduledCommandPublisher.cs
MichaelYoung1981: I started off with it on ICommandPublisher but they have different dependencies so it felt a bit odd once I'd done it.

PR #283, opened 2021-03-30
Ei 1030 minimum agreement version

David-B-Read:


src/SFA.DAS.EmployerIncentives.Data/ApprenticeshipIncentives/Models/ApprenticeshipIncentive.cs
```diff
@@ -41,6 +41,7 @@ public partial class ApprenticeshipIncentive
         [Column(TypeName = "nvarchar(50)")]
         public IncentiveStatus Status { get; set; }
 
+        public int MinimumAgreementVersion { get; set; }
```
ruslanss, line 44: should it be Nullable int?

src/SFA.DAS.EmployerIncentives.Domain/ApprenticeshipIncentives/ApprenticeshipIncentive.cs
```diff
@@ -31,10 +31,13 @@ public sealed class ApprenticeshipIncentive : AggregateRoot<Guid, Apprenticeship
         public IReadOnlyCollection<Payment> Payments => Model.PaymentModels.Map().ToList().AsReadOnly();
         public bool PausePayments => Model.PausePayments;
         public IReadOnlyCollection<ClawbackPayment> Clawbacks => Model.ClawbackPaymentModels.Map().ToList().AsReadOnly();
+        public int MinimumAgreementVersion => Model.MinimumAgreementVersion;
         private bool HasPaidEarnings => Model.PaymentModels.Any(p => p.PaidDate.HasValue);
-
+
```
ruslanss, line 36: why the whitespace?

src/SFA.DAS.EmployerIncentives.Database/Tables/ApprenticeshipIncentive.sql
```diff
@@ -18,7 +18,8 @@
     [SubmittedDate] DATETIME2 NULL, 
     [SubmittedByEmail] NVARCHAR(255) NULL,
 	[CourseName] NVARCHAR(126) NULL,
-	[Status] NVARCHAR(50) NULL
+	[Status] NVARCHAR(50) NULL,
+	[MinimumAgreementVersion] INT NULL
```
ruslanss, line 22: or maybe this field could have a calculated default value

src/SFA.DAS.EmployerIncentives.Data/ApprenticeshipIncentives/Models/ApprenticeshipIncentive.cs
David-B-Read: Good point!

src/SFA.DAS.EmployerIncentives.Data/ApprenticeshipIncentives/Models/ApprenticeshipIncentive.cs
```diff
@@ -41,6 +41,8 @@ public partial class ApprenticeshipIncentive
         [Column(TypeName = "nvarchar(50)")]
         public IncentiveStatus Status { get; set; }
 
+        public int? MinimumAgreementVersion { get; set; }
```
Stephek, line 44: Can you create a value object to hold this business logic.  e.g. an AgreementVersion value type with a static Calculate method.  Then in the incentive constructor  MinimumAgreementVersion = AgreementVersion.Calculate(DateTime startDate)

where Calculate returns a new AgreementVersion(int version) value type object?

That way the agreement logic is all contained in the value type and can be tested separately.

Also rather than expose the version expose the object e.g.

public AgreementVersion AgreementVersion { get; set; }

in fact does this need to be exposed yet?  If it's not used or required by anything else then should be kept private to the domain object.

Stephek (changes requested):

src/SFA.DAS.EmployerIncentives.Domain/ApprenticeshipIncentives/ApprenticeshipIncentive.cs
```diff
@@ -531,6 +532,19 @@ public void ValidateLearningData(Guid pendingPaymentId, Learner learner, Collect
             ValidateHasNoDataLocks(pendingPaymentId, learner, collectionPeriod);
             ValidateDaysInLearning(pendingPaymentId, learner, collectionPeriod);
         }
+        
+        private static int CalculateMinimumAgreementVersion(DateTime plannedStartDate)
```
Stephek, line 536: Delete ?

Stephek (changes requested):

src/SFA.DAS.EmployerIncentives.Data/ApprenticeshipIncentives/Map/DataExtensions.cs
```diff
@@ -69,7 +74,8 @@ internal static ApprenticeshipIncentiveModel Map(this ApprenticeshipIncentive en
                 PausePayments = entity.PausePayments,
                 SubmittedDate = entity.SubmittedDate,
                 SubmittedByEmail = entity.SubmittedByEmail,
-                Status = entity.Status
+                Status = entity.Status,
+                MinimumAgreementVersion = new AgreementVersion(entity.StartDate)
```
Stephek, line 78: This will recalculate each time on retrieval and will be wrong if there has been a start date change of circumstance.  The constructor needs to be passed the int?.  

Add a static factory method to the value type to create the initial agreement e.g.

public static Agreement Create(startdate)
{
   // existing logic
}

then this would become

MinimumAgreementVersion = entity.MinimumAgreementVersion.HasValue ? new AgreementVersion(version) : AgreementVersion.Create(entity.StartDate);

Then there isn't any conflict with the CoC story.

Stephek (changes requested):

Stephek (approved):

satya-sanikommu:
Testing completed

PR #290, opened 2021-04-22
EI-1075 Extend phase 1 eligibility period, create incentives configuration 

ruslanss:


src/SFA.DAS.EmployerIncentives.Domain.UnitTests/ValueObjects/WhenHasSignedRequiredAgreementVersion.cs
```diff
@@ -10,28 +10,63 @@ namespace SFA.DAS.EmployerIncentives.Domain.UnitTests.ValueObjects
     [TestFixture]
     public class WhenHasSignedRequiredAgreementVersion
     {
-        private List<IncentivePaymentProfile> _incentivePaymentProfiles;
+        private IncentivesConfiguration _config;
```
Stephek, line 13: Why change this?  Should still be  List<IncentivePaymentProfile>.

src/SFA.DAS.EmployerIncentives.Domain/ValueObjects/IncentivesConfiguration.cs
```diff
@@ -0,0 +1,65 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using SFA.DAS.EmployerIncentives.Abstractions.Domain;
+using SFA.DAS.EmployerIncentives.Domain.ApprenticeshipIncentives.Exceptions;
+using SFA.DAS.EmployerIncentives.Enums;
+
+namespace SFA.DAS.EmployerIncentives.Domain.ValueObjects
+{
+    public class IncentivesConfiguration
```
Stephek, line 10: Configuration is not a concern of the domain.  This is not a business object.   The IncentivePaymentProfile is the business object and already exists.  No need to add this object.

Stephek (changes requested):

src/SFA.DAS.EmployerIncentives.Domain/ValueObjects/IncentivesConfiguration.cs
ruslanss: Encapsulation, payment profile selection logic doesn't need to sit inside Incentive object. It is not its concern to filter our irrelevant payment configurations, don't you think? 🤷‍♂️

ruslanss: happy to rename this to make more sense?

src/SFA.DAS.EmployerIncentives.Domain.UnitTests/ValueObjects/WhenHasSignedRequiredAgreementVersion.cs
ruslanss: encapsulation

src/SFA.DAS.EmployerIncentives.Domain.UnitTests/ValueObjects/WhenCreatingTheIncentive.cs
```diff
@@ -11,25 +11,35 @@ namespace SFA.DAS.EmployerIncentives.Domain.UnitTests.ValueObjects
...
         [SetUp]
         public void SetUp()
         {
-            _incentivePaymentProfiles = new List<IncentivePaymentProfile>
+            var profiles = new List<IncentivePaymentProfile>
             {
-                new IncentivePaymentProfile(IncentiveType.TwentyFiveOrOverIncentive,
-                    new List<PaymentProfile>
-                        {new PaymentProfile(90, 1000), new PaymentProfile(365, 1000)}),
+                new IncentivePaymentProfile(
```
Stephek, line 21: The Phase should be a value object sine that is what is stored against the application and incentive.  And created as new Phase(string version);

Then the IIncentivePaymentProfileService can be extended with

IncentivePaymentProfile Get(Phase phase)

to return the IncentivePaymentProfile data for the phase.

Stephek (changes requested):

src/SFA.DAS.EmployerIncentives.Domain/ValueObjects/IncentivesConfiguration.cs
Stephek: It's not needed.  We have the IncentivePaymentProfile that encapsulates the structure and the Incentive object that uses that to provide business functionality.  The concern of the incentive object is to use and interpret the payment profiles to calculate the payments, eligibility and phase etc...   So this object is redundant since the business functionality is already encapsulated in the incentive object.   This is not adding any Business value so should be removed and the "helper" methods moved to the incentive object.

PR #292, opened 2021-04-22
EI-1075 996 extend phase 1 eligibility period

ruslanss:


src/SFA.DAS.EmployerIncentives.Domain/Interfaces/IIncentivePaymentProfilesService.cs
```diff
@@ -6,6 +6,6 @@ namespace SFA.DAS.EmployerIncentives.Domain.Interfaces
 {
     public interface IIncentivePaymentProfilesService
     {
-        Task<IEnumerable<IncentivePaymentProfile>> Get();
+        Task<IncentivesConfiguration> Get();
```
Stephek, line 9: This interface doesn't need to change.  The parent of the data is still an IncentivePaymentProfile object.  Extra data has been added to it and some moved around.

src/SFA.DAS.EmployerIncentives.Domain/ValueObjects/IncentivesConfiguration.cs
```diff
@@ -0,0 +1,65 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using SFA.DAS.EmployerIncentives.Abstractions.Domain;
+using SFA.DAS.EmployerIncentives.Domain.ApprenticeshipIncentives.Exceptions;
+using SFA.DAS.EmployerIncentives.Enums;
+
+namespace SFA.DAS.EmployerIncentives.Domain.ValueObjects
+{
+    public class IncentivesConfiguration
```
Stephek, line 10: This is not required and is not a business object.  There is already an IncentivePaymentProfile that does this.

Configuration is an infrastructure concern.

src/SFA.DAS.EmployerIncentives.Enums/IncentivePhase.cs
```diff
@@ -0,0 +1,9 @@
+﻿namespace SFA.DAS.EmployerIncentives.Enums
+{
+    public enum IncentivePhase : byte
```
Stephek, line 3: This should be a value object Phase with a constructor that accepts the phase string or byte (or whatever is to be stored in the db against the application or incentive)

src/SFA.DAS.EmployerIncentives.Queries/NewApprenticeIncentive/GetApplication/GetApplicationQueryHandler.cs
```diff
@@ -36,7 +36,7 @@ public async Task<GetApplicationResponse> Handle(GetApplicationRequest query, Ca
             return response;
         }
 
-        private bool IsNewAgreementRequired(IEnumerable<IncentiveApplicationApprenticeshipDto> applicationApprenticeships,  IEnumerable<IncentivePaymentProfile> paymentProfiles, int signedAgreementVersion)
+        private bool IsNewAgreementRequired(IEnumerable<IncentiveApplicationApprenticeshipDto> applicationApprenticeships,  IncentivesConfiguration paymentProfiles, int signedAgreementVersion)
```
Stephek, line 39: No need to change this interface.

src/SFA.DAS.EmployerIncentives.Domain/Interfaces/IIncentivePaymentProfilesService.cs
Stephek, line 9: Add a new Get overload  Task<IncentivePaymentProfile> Get(Phase phase) so the details can be loaded for the application/incentive if required.

Stephek (changes requested):

src/SFA.DAS.EmployerIncentives.Domain/ValueObjects/IncentivesConfiguration.cs
ruslanss: renamed

src/SFA.DAS.EmployerIncentives.Domain/Interfaces/IIncentivePaymentProfilesService.cs
ruslanss: we did not intend to store Phase Id in this user story

src/SFA.DAS.EmployerIncentives.Enums/IncentivePhase.cs
ruslanss: why?

Stephek (approved):

David-B-Read (approved):

PR #303, opened 2021-05-06
EI-1057 1078 earnings calculation for phase 2

Stephek:


src/SFA.DAS.EmployerIncentives.Data/ApprenticeApplicationDataRepository.cs
```diff
@@ -66,7 +66,8 @@ public async Task<List<ApprenticeApplicationDto>> GetList(long accountId, long a
                         InLearning = InLearning(data.learner),
                         PausePayments = data.incentive.PausePayments,
                         PaymentSent = data.firstPaymentSent != null,
-                        PaymentSentIsEstimated = IsPaymentEstimated(data.firstPaymentSent, _dateTimeService)
+                        PaymentSentIsEstimated = IsPaymentEstimated(data.firstPaymentSent, _dateTimeService),
+                        RequiresNewEmployerAgreement = !data.account.SignedAgreementVersion.HasValue || data.account.SignedAgreementVersion < data.incentive.MinimumAgreementVersion
```
ruslanss, line 70: should this logic be in the domain layer?

src/SFA.DAS.EmployerIncentives.UnitTests.Shared/Builders/Configuration/IncentivePaymentProfileBuilder.cs
```diff
@@ -0,0 +1,42 @@
+﻿using SFA.DAS.EmployerIncentives.Enums;
+using SFA.DAS.EmployerIncentives.Infrastructure.Configuration;
+using System.Collections.Generic;
+
+namespace SFA.DAS.EmployerIncentives.UnitTests.Shared.Builders.Configuration
+{
```
ruslanss, line 6: .cs filename needs renaming

src/SFA.DAS.EmployerIncentives.Data/ApprenticeApplicationDataRepository.cs
Stephek: good point.  That would need all the ApprenticeshipIncentives for a legal entity to be updated when the legal entity SignedAgreementVersion is changed and also storing a RequiresNewEmployerAgreement flag against the ApprenticeshipIncentive and updatting existing ones.  So quite a big change.  This way it's a read view of the model.  

Could make the above change as a tech debt ticket perhaps.  Speak to Mike?

Stephek: After speaking to Mike have created tech debt ticket 

https://skillsfundingagency.atlassian.net/browse/EI-1147

ruslanss (approved):

PR #312, opened 2021-05-14
EI-1061 Determine Phase 2 apprentices for select apprentices page

ruslanss:


MichaelYoung1981 (changes requested):
Those dates are phase 2 specific so should be on that object.  If we have EligibilityStartDates for each phase it'll prevent the need for another set of dates (which will be exactly the same) EI-1066.

MichaelYoung1981 (approved):

PR #322, opened 2021-05-28
Ei 1104 957 withdraw applications when legal entity removed

David-B-Read:


Stephek:
The RemoveLegalEntity method should raise a domain event and an event handler should publish the withdrawal commands. They are in different bounded contexts.

Stephek (approved):

src/SFA.DAS.EmployerIncentives.Events/Accounts/AccountLegalEntityRemovedHandler.cs
```diff
@@ -0,0 +1,42 @@
...
+using SFA.DAS.EmployerIncentives.Commands.Types.ApprenticeshipIncentive;
+using SFA.DAS.EmployerIncentives.Data.IncentiveApplication;
+using SFA.DAS.EmployerIncentives.Domain.Accounts.Events;
+
+namespace SFA.DAS.EmployerIncentives.Events.Accounts
+{
+    public class AccountLegalEntityRemovedHandler : IDomainEventHandler<AccountLegalEntityRemoved>
+    {
+        private readonly IIncentiveApplicationDataRepository _incentiveApplicationDataRepository;
+        private readonly ICommandDispatcher _commandDispatcher;
```
Stephek, line 15: Consider using the CommandPublisher.  That would put these on a queue for later processing by the Domain MessageHandlers and un-couple the two processes.

PR #325, opened 2021-06-10
EI-1125 BUG - When Resume COC happens - It shouldn't clawback valid First Payment paid

ruslanss:


src/SFA.DAS.EmployerIncentives.Domain/ApprenticeshipIncentives/ValueTypes/BreakInLearning.cs
```diff
@@ -13,14 +13,12 @@ public BreakInLearning(DateTime startDate)
 
         public DateTime StartDate { get; }
         public DateTime? EndDate { get; private set; }
+        public int Days => EndDate.HasValue ? (EndDate.Value - StartDate).Days : 0;
 
-        public void SetEndDate(DateTime dateTime)
+        public BreakInLearning SetEndDate(DateTime dateTime)
         {
-            if (dateTime.Date < StartDate)
```
Stephek, line 19: You should still have this since it's invalid data.  Or at least a <= check.

src/SFA.DAS.EmployerIncentives.Domain/ValueObjects/Incentive.cs
```diff
@@ -73,30 +71,33 @@ private static int AgeAtStartOfCourse(DateTime dateOfBirth, DateTime startDate)
...
 
             var paymentIndex = 0;
             foreach (var paymentProfile in paymentProfiles)
             {
-                payments.Add(new Payment(paymentProfile.AmountPayable, StartDate.AddDays(paymentProfile.DaysAfterApprenticeshipStart).AddDays(breakInLearningDayCount), _earningTypes[paymentIndex]));
+                payments.Add(new Payment(paymentProfile.AmountPayable, StartDate.AddDays(paymentProfile.DaysAfterApprenticeshipStart), _earningTypes[paymentIndex]));
                 paymentIndex++;
             }
 
+            foreach (var breakInLearning in breaksInLearning)
```
Stephek, line 86: Can this logic be moved to the Payment object that is applies to?.  i.e pass in the constructor.  Having an AddDays method on a value object (which should be immutable) seems wrong.

src/SFA.DAS.EmployerIncentives.Domain/ValueObjects/Payment.cs
```diff
@@ -19,12 +18,18 @@ public Payment(decimal amount, DateTime paymentDate, EarningType earningType)
             PaymentDate = paymentDate;
             EarningType = earningType;
         }
-      
+            
+        public void AddDays(int days)
```
Stephek, line 22: This breaks the Immutability of this value object.  Although we have done that elsewhere so need to fix those objects too.

Stephek (changes requested):

Stephek (approved):

PR #326, opened 2021-06-11
EI-1212-prevent-timeout-when-updating-payment-sent-status

Stephek:


ruslanss (dismissed):
since Payment isn't an aggregate root, don't think it should be updated in isolation. Should domain model be changed as well?

MichaelYoung1981:
> since Payment isn't an aggregate root, don't think it should be updated in isolation. Should domain model be changed as well?

I think you're probably right and it's hard to know why it wasn't done that way originally but I think it's outside of the scope of this change and it should be an additional tech debt ticket.

src/SFA.DAS.EmployerIncentives.Database/Tables/Payment.sql
```diff
@@ -13,6 +13,7 @@
 	[Amount] DECIMAL(9, 2) NOT NULL
 )
 GO
-CREATE INDEX IX_Payment_ApprenticeshipIncentiveId ON [incentives].[Payment] (ApprenticeshipIncentiveId)
+CREATE CLUSTERED INDEX IX_Payment_PaidDate ON [incentives].[Payment] (PaidDate)
```
MichaelYoung1981, line 16: I'm not convinced this index will be used so it'd be good to check the execution plan.  It certainly shouldn't be used when we have the IDs.

Stephek:
Payment definitely isn't an aggregate root.  The update is not being done through the domain.   It's not a domain update but part of the month end processing.  I assume it was implemented that way as an optimisation.  So yes a separate tech debt ticket is required to investigate and maybe to fix this.

Stephek:
> since Payment isn't an aggregate root, don't think it should be updated in isolation. Should domain model be changed as well?

Payment definitely isn't an aggregate root. The update is not being done through the domain. It's not a domain update but part of the month end processing. I assume it was implemented that way as an optimisation. So yes a separate tech debt ticket is required to investigate and maybe to fix this.

MichaelYoung1981 (approved):

PR #343, opened 2021-07-13
Ei 835 clawback payments

David-B-Read:


src/SFA.DAS.EmployerIncentives.Abstractions/DTOs/Queries/ClawbackStatusDto.cs
```diff
@@ -0,0 +1,10 @@
+﻿using System;
+
+namespace SFA.DAS.EmployerIncentives.Abstractions.DTOs.Queries
+{
+    public class ClawbackStatusDto
+    {
+        public decimal? ClawbackAmount { get; set; }
```
ruslanss, line 7: in what context are these going to be NULL?

src/SFA.DAS.EmployerIncentives.Abstractions/DTOs/Queries/ClawbackStatusDto.cs
```diff
@@ -0,0 +1,10 @@
...
+        public DateTime? ClawbackDate { get; set; }
```
ruslanss, line 8: in what context are these going to be `null`?
also Clawback prefix is implied and not needed imho 

```suggestion
        public decimal Amount { get; set; }
        public DateTime Date { get; set; }
```

David-B-Read: The thinking was that this line
https://github.com/SkillsFundingAgency/das-employer-incentives/blob/298dc4949e9dcec18f31fa5674b6fa079cd4dd45/src/SFA.DAS.EmployerIncentives.Data/ApprenticeApplicationDataRepository.cs#L38
would return an empty model if no clawback found, but it returns null, so have changed the dto accordingly.

ruslanss (approved):

PR #367, opened 2021-08-19
Ei 1350 use census date for learning stopped

David-B-Read:


src/SFA.DAS.EmployerIncentives.Domain/ValueObjects/CollectionCalendar.cs
```diff
@@ -93,6 +93,13 @@ public DateTime GetAcademicYearEndDate(string academicYearId)
             return academicYear.EndDate;
         }
 
+        public DateTime GetCensusDate()
```
Stephek, line 96: Can the caller not use the existing method to get the active period and read the property of the object?  Rather than have a whole load of specific functions just use the data already on the object.

Stephek (changes requested):

src/SFA.DAS.EmployerIncentives.Domain/ValueObjects/CollectionCalendar.cs
David-B-Read: Good point. I originally thought the census date method would be more complex than that, so split it out.
That's removed now

Stephek (approved):

PR #370, opened 2021-08-23
Exclude stopped and withdrawn applications.

MichaelYoung1981:


src/SFA.DAS.EmployerIncentives.Data/ApprenticeshipIncentives/ApprenticeshipIncentiveDataRepository.cs
```diff
@@ -30,7 +31,7 @@ public async Task<List<ApprenticeshipIncentiveModel>> FindApprenticeshipIncentiv
         {
             var collectionPeriods = _dbContext.CollectionPeriods.AsEnumerable();
 
-            var queryResults = _dbContext.ApprenticeshipIncentives.Where(x => x.PendingPayments.Count == 0);
+            var queryResults = _dbContext.ApprenticeshipIncentives.Where(x => x.Status != IncentiveStatus.Stopped && x.Status != IncentiveStatus.Withdrawn && x.PendingPayments.Count == 0);
```
Stephek, line 34: This is a very specific use case rule in the repository.  Can optional flags be added to the interface so the caller can decide what to include.

e.g.  FindApprenticeshipIncentivesWithoutPendingPayments(bool includePendingPayments = true, bool includeStopped = false, bool includeWithdrawn = false)

Or but those query options into a separate class  e.g. FindApprenticeshipIncentivesWithoutPendingPayments(FindApprenticeshipIncentivesFilter filter)  ?

MichaelYoung1981: I can understand your point but I don't think it is use case specific rule as I can't think of any scenario where you'd want to include stopped or withdrawn incentives in a list of incentives without pending payments.

Stephek: OK,  in that case shouldn't the same filter should be applied to the other data returned from the repository?

MichaelYoung1981: I think that would be safe but I'm not brave enough to see if my assumption is correct! :)  I've added parameters to allow withdrawn and stopped to be optionally included in the results.

Stephek (approved):

PR #372, opened 2021-08-24
EI-1295 mark earnings as calculated if pending payment records exist

David-B-Read:


src/SFA.DAS.EmployerIncentives.Commands/ApprenticeshipIncentive/CreateIncentive/CreateIncentiveCommandHandler.cs
```diff
@@ -7,28 +7,39 @@
...
             _apprenticeshipIncentiveDomainRepository = apprenticeshipIncentiveDomainRepository;
+            _commandPublisher = commandPublisher;
         }
 
         public async Task Handle(CreateIncentiveCommand command, CancellationToken cancellationToken = default)
         {
             var existing = await _apprenticeshipIncentiveDomainRepository.FindByApprenticeshipId(command.IncentiveApplicationApprenticeshipId);
             if (existing != null)
             {
+                if (existing.PendingPayments.Count > 0)
```
MichaelYoung1981, line 36: Could these be abstracted into a HasPendingPayment method on the existing object?

src/SFA.DAS.EmployerIncentives.Commands/ApprenticeshipIncentive/CreateIncentive/CreateIncentiveCommandHandler.cs
```diff
@@ -7,28 +7,39 @@
...
 
         public CreateIncentiveCommandHandler(
             IApprenticeshipIncentiveFactory apprenticeshipIncentiveFactory,
-            IApprenticeshipIncentiveDomainRepository apprenticeshipIncentiveDomainRepository)
+            IApprenticeshipIncentiveDomainRepository apprenticeshipIncentiveDomainRepository,
+            ICommandPublisher commandPublisher)
         {
             _apprenticeshipIncentiveFactory = apprenticeshipIncentiveFactory;
             _apprenticeshipIncentiveDomainRepository = apprenticeshipIncentiveDomainRepository;
+            _commandPublisher = commandPublisher;
```
MichaelYoung1981, line 28: You might be able to refactor this so the save repo method is still called as that'll publish the messages (I think!).  That way you can avoid the dependency.

Edit: this might be a terrible suggestion actually as you'd need to publish the EarningsCalculated domain event instead.

MichaelYoung1981 (changes requested):

src/SFA.DAS.EmployerIncentives.Commands/ApprenticeshipIncentive/CreateIncentive/CreateIncentiveCommandHandler.cs
```diff
@@ -7,28 +7,39 @@
...
         }
 
         public async Task Handle(CreateIncentiveCommand command, CancellationToken cancellationToken = default)
         {
             var existing = await _apprenticeshipIncentiveDomainRepository.FindByApprenticeshipId(command.IncentiveApplicationApprenticeshipId);
             if (existing != null)
             {
+                if (existing.PendingPayments.Count > 0)
+                {
+                    var completeEarningCalculationCommand = new CompleteEarningsCalculationCommand(existing.Account.Id,
```
Stephek, line 38: This is breaking all our patterns.  Shouldn't the resilience check be looking for these scenarios/issues and publishing the CompleteEarningsCalculationCommand ?

Stephek (changes requested):
Need to discuss this.

src/SFA.DAS.EmployerIncentives.Commands/ApprenticeshipIncentive/CreateIncentive/CreateIncentiveCommandHandler.cs
MichaelYoung1981: > This is breaking all our patterns. Shouldn't the resilience check be looking for these scenarios/issues and publishing the CompleteEarningsCalculationCommand ?
> 
> Or better handlethis in the SubmittedHandler event handler which can look up the incentive and Publish the CompleteEarningsCalculationCommand if needed else submit the CreateIncentiveCommand.

I'm not sure the SubmittedHandler would be better as it'd be crossing bounded contexts if it needs to get the ApprenticeshipIncentive.  Having a separate step in the resilience check would probably make sense tho.

Stephek: Yep, the purpose of the resilience check is to fix corrupt data so makes sense to add a step there to do just that.

src/SFA.DAS.EmployerIncentives.Commands/EarningsResilienceCheck/EarningsResilienceApplicationsCheckCommandHandler.cs
```diff
@@ -29,6 +33,12 @@ public async Task Handle(EarningsResilienceApplicationsCheckCommand command, Can
                 {
                     tasks.Add(_domainEventDispatcher.Send(domainEvent, cancellationToken));
                 }
+                foreach (var apprenticeship in application.Apprenticeships)
```
Stephek, line 36: Should this only apply to incentives where PendingPayments.Count > 0 ?
i.e. should it be a separate resilience check (separate command handler) for that scenario since it's a different use case? 

So look up incentives with pending payments and for those that don't have apprentices with EarningsCalculated true then fire this event?

Stephek (changes requested):

src/SFA.DAS.EmployerIncentives.Commands/EarningsResilienceCheck/EarningsResilienceApplicationsCheckCommandHandler.cs
David-B-Read: I think that would have to be in a separate handler, the earnings resilience handler works with applications, not the incentives, so would be across different bounded contexts

Stephek (approved):

David-B-Read:
Closing because no new applications will be made for Employer Incentives. Created a datafix script to update the remaining 52 records to exclude them from being re-processed by the Earnings Resilience check

PR #373, opened 2021-08-25
EI-1356 rename learner table column to SuccessfulLearnerMatchExecution

David-B-Read:


Stephek (commented):
Wasn't this just a db column change?

PR #383, opened 2021-09-03
EI-1420 add new agreement version and Phase2 eligibility check

David-B-Read:


src/SFA.DAS.EmployerIncentives.Domain/ValueObjects/Incentive.cs
```diff
@@ -179,6 +181,12 @@ public class Phase2Incentive : Incentive
 
         public static int MinimumAgreementVersion() => 6;
```
ruslanss, line 182: are we keeping this?

src/SFA.DAS.EmployerIncentives.Domain/ValueObjects/Incentive.cs
```diff
@@ -169,6 +169,8 @@ public class Phase2Incentive : Incentive
         {
         }
 
+        public static int MinimumPhase2AgreementVersion => 6;
```
ruslanss, line 172: why the "Phase2" when it's implied?

David-B-Read: You can't have a method and a property with the same name, so had to rename it. Changed to use the overloaded method.

src/SFA.DAS.EmployerIncentives.Domain/ValueObjects/Incentive.cs
David-B-Read: Yep

ruslanss (approved):

PR #401, opened 2021-10-01
EI-1480 store last learner match JSON response even if unable to process

David-B-Read:


David-B-Read:
Closing as the database transaction is rolled back when an exception is thrown, so the updated JSON wouldn't get stored. This change throws an exception for an invalid response which would include the JSON in the logs: https://github.com/SkillsFundingAgency/das-employer-incentives/pull/448

PR #424, opened 2021-11-11
Ei 1578 correct days in learning validation

MichaelYoung1981:


David-B-Read (approved):

src/SFA.DAS.EmployerIncentives.Domain/ValueObjects/Incentive.cs
```diff
@@ -58,6 +58,27 @@ public static bool EmployerStartDateIsEligible(Apprenticeship apprenticeship)
             return false;
         }
 
+        public static List<PaymentProfile> GetPaymentProfiles(Phase phase, DateTime dateOfBirth, DateTime startDate, IEnumerable<IncentivePaymentProfile> incentivePaymentProfiles)
```
Stephek, line 61: Rather than a new static method here can it not go on the Incentive abstract base class?

protected int DaysAfterApprenticeshipStart (); 

?

there's already a Static Incentive.Create method used by the ApprenticeshipIncentive e.g.

var incentive = Incentive.Create(this, paymentProfiles);

This incentive returned has the Phase, DateOfBirth, StartDate and PaymentProfiles which is everything required to return the DaysAfterApprenticeshipStart

Stephek (changes requested):

Stephek (approved):

PR #437, opened 2021-12-02
Ei 1596 force new employment check

David-B-Read:


src/SFA.DAS.EmployerIncentives.Commands/ApprenticeshipIncentive/EmploymentCheck/RefreshEmploymentChecksCommandHandler.cs
```diff
@@ -0,0 +1,30 @@
...
+            _incentiveDomainRepository = incentiveDomainRepository;
+            _commandPublisher = commandPublisher;
+        }
+        public async Task Handle(RefreshEmploymentChecksCommand command, CancellationToken cancellationToken = default(CancellationToken))
+        {
+            var incentivesWithLearning = await _incentiveDomainRepository.FindIncentivesWithLearningFound();
+
+            foreach(var incentive in incentivesWithLearning)
+            {
+                await _commandPublisher.Publish(new SendEmploymentCheckRequestsCommand(incentive.Id));
```
MichaelYoung1981, line 25: I don't think does what's needed as that won't create the employment check records if they don't exist or replace them if they do exist will it?

src/SFA.DAS.EmployerIncentives.Api/Controllers/EmploymentCheckController.cs
```diff
@@ -6,14 +6,21 @@
...
+using SFA.DAS.EmployerIncentives.Domain.Interfaces;
 
 namespace SFA.DAS.EmployerIncentives.Api.Controllers
 {
     [Route("api/[controller]")]
     [ApiController]
     public class EmploymentCheckController : ApiCommandControllerBase
     {
-        public EmploymentCheckController(ICommandDispatcher commandDispatcher) : base(commandDispatcher) { }
+        private readonly ICollectionCalendarService _collectionCalendarService;
```
MichaelYoung1981, line 17: I don't think this should be injected into the controller, maybe move the logic to the command handler?

MichaelYoung1981 (changes requested):

src/SFA.DAS.EmployerIncentives.Api/Controllers/EmploymentCheckController.cs
```diff
@@ -29,6 +32,15 @@ public async Task<IActionResult> Update([FromBody] UpdateEmploymentCheckRequest
             return Ok($"/employmentchecks/{updateRequest.CorrelationId}");
         }
 
+        [HttpPut("/employmentchecks/refresh")]
```
Stephek, line 35: This is not a good REST endpoint.  Please use the existing Jobs endpoint and create a new RefreshEmploymentChecks job since that is what is required by the caller.

src/SFA.DAS.EmployerIncentives.Commands/ApprenticeshipIncentive/EmploymentCheck/RefreshEmploymentChecksCommand.cs
```diff
@@ -0,0 +1,22 @@
+﻿using SFA.DAS.EmployerIncentives.Abstractions.Commands;
+using SFA.DAS.EmployerIncentives.Abstractions.Logging;
+
+namespace SFA.DAS.EmployerIncentives.Commands.ApprenticeshipIncentive.EmploymentCheck
+{
+    public class RefreshEmploymentChecksCommand : DomainCommand, ILogWriter
```
Stephek, line 6: Implement the IPeriodEndIncompatible to delay or prevent the command running.  You may need to improve the interface if the requirement is to abandon the command from running rather than just delay it.

src/SFA.DAS.EmployerIncentives.Commands/ApprenticeshipIncentive/EmploymentCheck/RefreshEmploymentChecksCommandHandler.cs
```diff
@@ -0,0 +1,55 @@
...
+                                                     ICommandPublisher commandPublisher,
+                                                     ICollectionCalendarService collectionCalendarService)
+        {
+            _incentiveDomainRepository = incentiveDomainRepository;
+            _commandPublisher = commandPublisher;
+            _collectionCalendarService = collectionCalendarService;
+        }
+        public async Task Handle(RefreshEmploymentChecksCommand command, CancellationToken cancellationToken = default(CancellationToken))
+        {
+            if (await ActivePeriodInProgress())
```
Stephek, line 27: Use the IPeriodEndIncompatible on the command so that the decorator implements this functionality so this is not duplicated and littered through out the code base but is maintained in one place.

Stephek (changes requested):

src/SFA.DAS.EmployerIncentives.Commands/ApprenticeshipIncentive/EmploymentCheck/RefreshEmploymentChecksCommandHandler.cs
```diff
@@ -0,0 +1,38 @@
...
+
+            foreach(var incentive in incentivesWithLearning)
+            {
+                if (!incentive.HasEmploymentChecks)
+                {
+                    incentive.AddEmploymentChecks();
+                    await _incentiveDomainRepository.Save(incentive);
+                }
+
+                await _commandPublisher.Publish(new SendEmploymentCheckRequestsCommand(incentive.Id));
```
Stephek, line 33: When AddEmploymentChecks is called the EmploymentChecksCreated event is raised.  The EmploymentChecksCreatedHandler already handles that event and publishes the SendEmploymentCheckRequestsCommand.  So this line can be removed completely.

Stephek (changes requested):

src/SFA.DAS.EmployerIncentives.Commands/ApprenticeshipIncentive/EmploymentCheck/RefreshEmploymentChecksCommandHandler.cs
David-B-Read: I think it's still needed for when the incentive already has employment checks that need refreshing, so will add an else condition

Stephek: But that AddEmploymentChecks method clears them and replaces them doesn't it ?

```
 public void AddEmploymentChecks()
        {
            Model.EmploymentCheckModels.Clear();
```

David-B-Read: So use AddEmploymentChecks regardless of whether the incentive already has checks.

Stephek: Isn't the use case to **force** a **new** employment check for all apprenticeships that have a learner record and haven't been withdrawn?  Double check with Mike but I think existing checks are to be replaced.  This may be required if the rules change and all existing incentives have to be re-checked.

MichaelYoung1981: Yep, they all need to be rechecked for this process.

src/SFA.DAS.EmployerIncentives.Domain/ApprenticeshipIncentives/ApprenticeshipIncentive.cs
```diff
@@ -712,7 +713,7 @@ private void RequestEmploymentChecks(bool? isInLearning)
             AddEmploymentChecks();
         }
 
-        private void AddEmploymentChecks()
+        public void AddEmploymentChecks()
         {
             Model.EmploymentCheckModels.Clear();
```
Stephek, line 718: I think we should be checking that the model status is not withdrawn here.

Stephek (changes requested):

Stephek (approved):

MichaelYoung1981 (approved):

PR #439, opened 2021-12-03
EI 1584 only store later changes and add updated audit date

Stephek:


David-B-Read (approved):

Stephek:
Merged into parent branch

PR #455, opened 2022-01-24
EI-1611-add-validation-override-support

Stephek:


src/SFA.DAS.EmployerIncentives.Api.AcceptanceTests/Features/ValidationOverride.feature
```diff
@@ -0,0 +1,44 @@
...
+	I want to be able to override payment validations
+
+Scenario Outline: A payment validation override is created
+	Given an apprenticeship incentive has a <validationType> validation
+	When the validation override request is received
+	Then the validation override <validationType> is stored against the apprenticeship incentive
+
+Examples:
+    | validationType                  |
+    | HasBankDetails                  |
```
David-B-Read, line 18: According to the JIRA ticket, we shouldn't support all validation types. Overriding HasBankDetails would cause a failure in the payment process because the employer couldn't be paid

src/SFA.DAS.EmployerIncentives.Api/Types/ValidationType.cs
```diff
@@ -0,0 +1,17 @@
+﻿namespace SFA.DAS.EmployerIncentives.Api.Types
+{
+    public enum ValidationType
+    {
```
David-B-Read, line 4: Think this enum should just include:
HasDaysInLearning
HasNoDataLocks
IsInLearning
EmployedBeforeSchemeStarted
EmployedAtStartOfApprenticeship

Anything else should throw an error

src/SFA.DAS.EmployerIncentives.Api.AcceptanceTests/Features/ValidationOverride.feature
Stephek: The allowable overrides are restricted in the Functions so HasBankDetails etc.. should never get called.  I didn't duplicated that here so that could be easily changed.

Stephek: Updated

src/SFA.DAS.EmployerIncentives.Api/Types/ValidationType.cs
Stephek: Updated

David-B-Read (approved):

PR #456, opened 2022-01-24
Ei 1661 employment check not run

David-B-Read:


MichaelYoung1981 (approved):
I've got concerns about the performance of the query being generated to fulfil the request so maybe we should create a tech debt ticket to address that?

David-B-Read:
> I've got concerns about the performance of the query being generated to fulfil the request so maybe we should create a tech debt ticket to address that?

Good idea, it is becoming a bit unwieldy - perhaps replace the EF query with a SQL view?

PR #463, opened 2022-02-07
Ei 1669 reinstate withdrawn application

MichaelYoung1981:


src/SFA.DAS.EmployerIncentives.Api/Controllers/WithdrawalCommandController.cs
```diff
@@ -49,8 +48,21 @@ public async Task<IActionResult> WithdrawalIncentiveApplication([FromBody] Withd
             }
             else
             {
-                return BadRequest(new { Error = "Invalid WithdrawalType of {request.WithdrawalType} passed in" });
+                return BadRequest(new {Error = "Invalid WithdrawalType of {request.WithdrawalType} passed in"});
             }
-        }       
+        }
+
+        [HttpPost("/withdrawals/reinstate")]
```
Stephek, line 55: The REST guidelines are that resource URI's should be based on nouns not verbs (https://docs.microsoft.com/en-us/azure/architecture/best-practices/api-design)  so this would be better named WithDrawalReinstatements or just Reinstatements resource (with a ReinstatementType of Withdrawal on the request).

David-B-Read (approved):

PR #472, opened 2022-02-21
Ei 1679 block account legal entities for payment

David-B-Read:


src/SFA.DAS.EmployerIncentives.Api/Controllers/AccountCommandController.cs
```diff
@@ -60,5 +61,20 @@ public async Task<IActionResult> AddEmployerVendorId([FromRoute] string hashedLe
             return NoContent();
         }
 
+        [HttpPatch("/legalentities/blockpayments")]
```
Stephek, line 64: Should be a noun.  Also either /blockedpayments or /legalentities/<id>/blockedpayments

David-B-Read: Updated to /blockedpayments as there are multiple IDs in the JSON payload

ruslanss (approved):

PR #480, opened 2022-03-15
Ei 1709 ytd validation figures

David-B-Read:


src/SFA.DAS.EmployerIncentives.Database/Scripts/Generate.MonthEndMetrics.sql
```diff
@@ -33,6 +33,32 @@ SELECT PaymentYear as [(Paste into cell A6) PaymentYear],PaymentPeriod, count(*)
...
+	FROM incentives.Payment
+	WHERE PaymentYear = @AcademicYear
+	AND PaidDate IS NOT NULL
+	UNION ALL
+	SELECT SUM(Amount)
+	FROM incentives.PendingPayment
+	WHERE Id 
+	IN
+	(
+	SELECT DISTINCT PendingPaymentId FROM incentives.PendingPaymentValidationResult
```
Stephek, line 56: Shouldn't this be only for failed validation results?

David-B-Read: The query takes the value of all the pending payments for the academic year that have been validated, and then subtracts the value of all the ones that have passed validation and been paid. The remainder are the ones that have been validated but not paid, so no filter required, I think

Stephek: OK.  Can you add a comment to the source code s it's clear in the future?

David-B-Read: Done

PR #481, opened 2022-03-17
EI-1690 calculate earnings if learning stopped date after due date

David-B-Read:


Stephek (approved):

src/SFA.DAS.EmployerIncentives.Domain/ApprenticeshipIncentives/ApprenticeshipIncentive.cs
```diff
@@ -335,6 +335,19 @@ private void StartBreakInLearning(DateTime startDate)
                     Model.Id,
                     learningStoppedStatus.DateResumed.Value));
             }
+            else if (Model.Status == IncentiveStatus.Stopped && learningStoppedStatus.LearningStopped && PendingPayments.Count == 0)
```
MichaelYoung1981, line 338: I'm not sure if this works...

1. There may already be a first earning and the change in stop date could change whether the second earning is due
2. If the earnings are recalculated then, from memory, the stop date isn't factored in so you could well end up with earnings after the new stop date

MichaelYoung1981 (approved):

PR #490, opened 2022-04-05
EI-1744 add index for querying incentives by account legal entity id

David-B-Read:


David-B-Read:
Note that ULN index is added in https://github.com/SkillsFundingAgency/das-employer-incentives/pull/488

MichaelYoung1981 (approved):

PR #522, opened 2022-07-29
EI-1654-migrate_to_net6-wip

Stephek:


azure-pipelines.yml
```diff
@@ -195,13 +174,41 @@ stages:
...
       displayName: 'Publish Artifact: drop'
       inputs:
         targetPath: $(build.artifactstagingdirectory)/publish
         artifactName: drop
 
+  - job: 'BuildDacpac'
+    pool:
+      name: 'DAS - Continuous Integration'
+      demands:
+      - Agent.OS -equals Windows_NT
```
David-B-Read, line 188: Is this correct? I thought the build was running on Linux build agents now

src/SFA.DAS.EmployerIncentives.TestHelpers/Images/Dockerfile
```diff
@@ -0,0 +1,35 @@
+﻿FROM mcr.microsoft.com/mssql/server:2019-latest
+ENV ACCEPT_EULA=Y
+ENV SA_PASSWORD=Pa55word!
```
David-B-Read, line 3: Password in plain text?

src/SFA.DAS.EmployerIncentives.TestHelpers/Types/DatabaseInfo.cs
```diff
@@ -0,0 +1,14 @@
+﻿namespace SFA.DAS.EmployerIncentives.TestHelpers.Types
```
David-B-Read, line 1: Can this file be removed?

src/SFA.DAS.EmployerIncentives.TestHelpers/Images/Dockerfile
Stephek: This is just for the test db.  Could try and replace it with the test Instance Id I suppose.  Easier for the devs if it's the same value though.

azure-pipelines.yml
Stephek: That's just for the dacpac build part which is still built on the NT box as per the devops standard/suggestion.  This approach was provided by devops.

src/SFA.DAS.EmployerIncentives.TestHelpers/Types/DatabaseInfo.cs
Stephek: Done

David-B-Read (approved):

azure-pipelines.yml
```diff
@@ -24,18 +24,18 @@ resources:
   - repository: das-platform-building-blocks
     type: github
     name: SkillsFundingAgency/das-platform-building-blocks
-    ref: refs/tags/2.1.0
+    ref: refs/tags/0.4.63
```
HussainsGitHub, line 27: Could this tag be updated to be 2.1.4

HussainsGitHub (approved):

markusokafor (approved):

PR #531, opened 2022-08-16
EI-1898 EI-1901 add employed at 365 days check to metrics report

David-B-Read:


src/SFA.DAS.EmployerIncentives.Database/Scripts/Generate.MonthEndMetrics.sql
```diff
@@ -265,6 +272,11 @@ select max(ppv.periodnumber) MaxPeriod,
 	ELSE max((case when ppv.periodnumber < 10 and ppv.paymentyear <= 2122 then 1 else iif(step='BlockedForPayments' and result=1,1,0) end)) 
 	END
 	as BlockedForPayments,  
+  CASE OverrideResult
+  WHEN 1 THEN 1
+  ELSE max((case when ppv.paymentyear <= 2223 then 1 else iif(step='EmployedAt365Days' and result=1,1,0) end))
```
David-B-Read, line 277: Note: the script assumes that the first execution of the 365 day employment check will be in period R01 academic year 2223. If this is not the case, this line will need to be updated to include the period number.

Stephek (approved):

PR #539, opened 2022-09-05
EI-1894 handle error codes for previous checks

David-B-Read:


src/SFA.DAS.EmployerIncentives.Domain/ApprenticeshipIncentives/ApprenticeshipIncentive.cs
```diff
@@ -1012,7 +1013,23 @@ private bool HasSuccessfulChecks(IList<EmploymentCheckType> checkTypes)
 
         private bool HasEmploymentCheck(EmploymentCheckType employmentCheckType)
```
Stephek, line 1014: Should this be renamed HasSuccessfulEmploymentCheck?

David-B-Read: Renamed to CanRefreshEmploymentCheck as this is under the following conditions: 
- a check result has been returned
- a check result isn't true or false but an error code was returned indicating that the check needs to be re-run

Stephek (approved):

PR #540, opened 2022-09-07
Ei 1913 reinstate when multiple applications

David-B-Read:


src/SFA.DAS.EmployerIncentives.Commands/Withdrawals/ReinstateWithdrawal/ReinstateWithdrawalCommandHandler.cs
```diff
@@ -1,52 +1,62 @@
...
         }
 
         public async Task Handle(ReinstateWithdrawalCommand command, CancellationToken cancellationToken = default)
         {
             var applications = await _domainRepository.Find(command.AccountLegalEntityId, command.ULN);
-            if(!applications.Any())
+            if(applications == null || !applications.Any())
             {
-                throw new WithdrawalException($"Unable to handle reinstate withdrawal command.  No matching incentive applications found for {command}");
+                throw new WithdrawalException($"Unable to handle reinstate withdrawal command.  No matching incentive applications found for ULN {command.ULN}");
```
Stephek, line 28: Should we be logging Uln's ?  Isn't that PII data?

Stephek (changes requested):

src/SFA.DAS.EmployerIncentives.Commands/Withdrawals/ReinstateWithdrawal/ReinstateWithdrawalCommandHandler.cs
David-B-Read: Good point, they've been removed now

Stephek (approved):

PR #541, opened 2022-09-08
Ei 1940 split queries for apprenticeshipincentive

David-B-Read:


src/SFA.DAS.EmployerIncentives.Data/ApprenticeshipIncentives/ApprenticeshipIncentiveDataRepository.cs
```diff
@@ -151,6 +129,34 @@ public async Task Delete(ApprenticeshipIncentiveModel apprenticeshipIncentive)
...
+        }
+
+        private async Task PopulateApprenticeshipIncentive(ApprenticeshipIncentive apprenticeshipIncentive)
+        {
+            apprenticeshipIncentive.PendingPayments = await _dbContext.PendingPayments.Where(x => x.ApprenticeshipIncentiveId == apprenticeshipIncentive.Id).ToListAsync();
+            foreach (var pendingPayment in apprenticeshipIncentive.PendingPayments)
+            {
+                pendingPayment.ValidationResults = await _dbContext.PendingPaymentValidationResults.Where(x => x.PendingPaymentId == pendingPayment.Id).ToListAsync();
+            }
+            apprenticeshipIncentive.Payments = await _dbContext.Payments.Where(x => x.ApprenticeshipIncentiveId == apprenticeshipIncentive.Id).ToListAsync();
```
MichaelYoung1981, line 152: Would there be any harm in doing these gets concurrently? As long as multiple active record sets is enabled it should be quicker?

src/SFA.DAS.EmployerIncentives.Data/ApprenticeshipIncentives/ApprenticeshipIncentiveDataRepository.cs
```diff
@@ -75,9 +62,9 @@ public async Task<List<ApprenticeshipIncentiveModel>> FindApprenticeshipIncentiv
                 .Include(x => x.BreakInLearnings)
```
hardik-desai-kcl, line 62: did this method not work with the the new Get?

src/SFA.DAS.EmployerIncentives.Data/ApprenticeshipIncentives/ApprenticeshipIncentiveDataRepository.cs
```diff
@@ -151,6 +129,34 @@ public async Task Delete(ApprenticeshipIncentiveModel apprenticeshipIncentive)
...
+                return apprenticeshipIncentive?.Map(_dbContext.CollectionPeriods.AsEnumerable());
+            }
+
+            return null;
+        }
+
+        private async Task PopulateApprenticeshipIncentive(ApprenticeshipIncentive apprenticeshipIncentive)
+        {
+            apprenticeshipIncentive.PendingPayments = await _dbContext.PendingPayments.Where(x => x.ApprenticeshipIncentiveId == apprenticeshipIncentive.Id).ToListAsync();
+            foreach (var pendingPayment in apprenticeshipIncentive.PendingPayments)
```
hardik-desai-kcl, line 148: foreach loop might be slower and more expensive i.e. number of round trip to SQL server etc compare to doing include, alternatively use an "in" statement (IN statement does have a limit of 2000 or something so I would check actual data and do max per apprenticeshipIncentive and based on that chose one or the other)

src/SFA.DAS.EmployerIncentives.Data/ApprenticeshipIncentives/ApprenticeshipIncentiveDataRepository.cs
```diff
@@ -151,6 +129,34 @@ public async Task Delete(ApprenticeshipIncentiveModel apprenticeshipIncentive)
...
+            {
+                pendingPayment.ValidationResults = await _dbContext.PendingPaymentValidationResults.Where(x => x.PendingPaymentId == pendingPayment.Id).ToListAsync();
+            }
+            apprenticeshipIncentive.Payments = await _dbContext.Payments.Where(x => x.ApprenticeshipIncentiveId == apprenticeshipIncentive.Id).ToListAsync();
+            apprenticeshipIncentive.ClawbackPayments = await _dbContext.ClawbackPayments.Where(x => x.ApprenticeshipIncentiveId == apprenticeshipIncentive.Id).ToListAsync();
+            apprenticeshipIncentive.BreakInLearnings = await _dbContext.BreakInLearnings.Where(x => x.ApprenticeshipIncentiveId == apprenticeshipIncentive.Id).ToListAsync();
+            apprenticeshipIncentive.EmploymentChecks = await _dbContext.EmploymentChecks.Where(x => x.ApprenticeshipIncentiveId == apprenticeshipIncentive.Id).ToListAsync();
+            apprenticeshipIncentive.ValidationOverrides = await _dbContext.ValidationOverrides.Where(x => x.ApprenticeshipIncentiveId == apprenticeshipIncentive.Id).ToListAsync();
```
hardik-desai-kcl, line 156: one way to improve this method is to run all queries in parallel but just adding AwaitAll will not be enough you need to create datacontext factory to create a new copy of DBContect otherwise using same instance of context means they will probably run one after the other

hardik-desai-kcl: https://stackoverflow.com/questions/41749896/ef-6-how-to-correctly-perform-parallel-queries

src/SFA.DAS.EmployerIncentives.Data/ApprenticeshipIncentives/ApprenticeshipIncentiveDataRepository.cs
hardik-desai-kcl: @MichaelYoung1981 it will need DBContext Factory, its still doable not a big deal
https://stackoverflow.com/questions/41749896/ef-6-how-to-correctly-perform-parallel-queries

src/SFA.DAS.EmployerIncentives.Data/ApprenticeshipIncentives/ApprenticeshipIncentiveDataRepository.cs
hardik-desai-kcl: https://github.com/SkillsFundingAgency/das-payments-V2-matched-learner/blob/master/src/SFA.DAS.Payments.MatchedLearner.Data/Contexts/MatchedLearnerDataContextFactory.cs

src/SFA.DAS.EmployerIncentives.Data/ApprenticeshipIncentives/ApprenticeshipIncentiveDataRepository.cs
David-B-Read: No, unfortunately not. The new Get does not allow changes to be tracked in the EF model in the unit of work.
Specifically, the acceptance test `MultipleValidationOverrideRequestsAreHandled` in `ValidationOverrideFeature` fails because two commands make updates to the model, and then the changes are persisted once the transaction is complete. When reading the child entities separately, the changes are not tracked and are overwritten with the current state of the database records.

src/SFA.DAS.EmployerIncentives.Data/ApprenticeshipIncentives/ApprenticeshipIncentiveDataRepository.cs
David-B-Read: Typically there are 2 pending payments per incentive, there can be more than that if payments are clawed back and recalculated, or less if earnings removed, but those are outliers

src/SFA.DAS.EmployerIncentives.Data/ApprenticeshipIncentives/ApprenticeshipIncentiveDataRepository.cs
```diff
@@ -151,6 +129,34 @@ public async Task Delete(ApprenticeshipIncentiveModel apprenticeshipIncentive)
             }
         }
 
+        private async Task<ApprenticeshipIncentiveModel> Get(Expression<Func<ApprenticeshipIncentive, bool>> predicate)
+        {
+            var apprenticeshipIncentive = await _dbContext.ApprenticeshipIncentives.FirstOrDefaultAsync(predicate);
```
Stephek, line 134: You could pull back the PendingPayments and ValidationResults at the same time as the ApprenticeshipIncentives here.    the query should still be efficient since its parent-table->child-table->child-table .  Is the subsequent includes that cause the problems.

src/SFA.DAS.EmployerIncentives.Data/ApprenticeshipIncentives/ApprenticeshipIncentiveDataRepository.cs
```diff
@@ -151,6 +129,34 @@ public async Task Delete(ApprenticeshipIncentiveModel apprenticeshipIncentive)
...
+
+            return null;
+        }
+
+        private async Task PopulateApprenticeshipIncentive(ApprenticeshipIncentive apprenticeshipIncentive)
+        {
+            apprenticeshipIncentive.PendingPayments = await _dbContext.PendingPayments.Where(x => x.ApprenticeshipIncentiveId == apprenticeshipIncentive.Id).ToListAsync();
+            foreach (var pendingPayment in apprenticeshipIncentive.PendingPayments)
+            {
+                pendingPayment.ValidationResults = await _dbContext.PendingPaymentValidationResults.Where(x => x.PendingPaymentId == pendingPayment.Id).ToListAsync();
```
Stephek, line 150: There's no need for this for each loop. Not all records need to be singularly fetched.   The PendingPaymentValidationResults can come back with the PendingPayments so just use the Include() in this case.

src/SFA.DAS.EmployerIncentives.Data/ApprenticeshipIncentives/ApprenticeshipIncentiveDataRepository.cs
David-B-Read: Have tried that, unfortunately the queries can't be run in a separate data context as it prevents the changes from being tracked and persisted in the Unit of Work

src/SFA.DAS.EmployerIncentives.Data/ApprenticeshipIncentives/ApprenticeshipIncentiveDataRepository.cs
David-B-Read: This has been changed to fetch the pending payments and validation results in the initial fetch, as per Kevin's suggestion

src/SFA.DAS.EmployerIncentives.Data/ApprenticeshipIncentives/ApprenticeshipIncentiveDataRepository.cs
David-B-Read: Updated

src/SFA.DAS.EmployerIncentives.Data/ApprenticeshipIncentives/ApprenticeshipIncentiveDataRepository.cs
David-B-Read: Updated

src/SFA.DAS.EmployerIncentives.Data/ApprenticeshipIncentives/ApprenticeshipIncentiveDataRepository.cs
David-B-Read: See comments about the separate database contexts and EF unit of work

src/SFA.DAS.EmployerIncentives.Data/ApprenticeshipIncentives/ApprenticeshipIncentiveDataRepository.cs
Stephek: +1 for the acceptance tests

src/SFA.DAS.EmployerIncentives.Data/ApprenticeshipIncentives/ApprenticeshipIncentiveDataRepository.cs
Stephek: Is it not possible to do something like this ?

`
var paymentsTask = _dbContext.Payments.Where(x => x.ApprenticeshipIncentiveId == apprenticeshipIncentive.Id).AsNoTracking().ToListAsync();
            var clawbackPaymentsTask = _dbContext.Payments.Where(x => x.ApprenticeshipIncentiveId == apprenticeshipIncentive.Id).AsNoTracking().ToListAsync();

            await Task.WhenAll(paymentsTask, clawbackPaymentsTask);

            _dbContext.AttachRange(paymentsTask.Result);
            _dbContext.AttachRange(clawbackPaymentsTask.Result);
`

David-B-Read: Have just tried that - get the following exception when running the first `_dbContext.AttachRange()` :

`System.InvalidOperationException : The instance of entity type 'Payment' cannot be tracked because another instance with the same key value for {'Id'} is already being tracked. When attaching existing entities, ensure that only one entity instance with a given key value is attached. Consider using 'DbContextOptionsBuilder.EnableSensitiveDataLogging' to see the conflicting key values.`

hardik-desai-kcl (approved):

PR #549, opened 2022-10-04
EI-1942 calculate learning periods with breaks across academic years

David-B-Read:


src/SFA.DAS.EmployerIncentives.Commands/Services/LearnerMatchApi/LearnerDataExtensions.cs
```diff
@@ -171,8 +171,15 @@ private static IList<LearningPeriod> MergeLearningPeriods(IList<LearningPeriod>
         {
             for (var i = 0; i < periods.Count-1; i++)
             {
-                var gap = (periods[i + 1].StartDate - periods[i].EndDate).Days;
-                if (gap <= 1)
+                // merge learning periods if:
+                // periods are adjacent
+                // periods overlap
```
Stephek, line 176: Can you refactor the code so that they read as per the comments  e.g. if(PeriodsAreAdjacent() || PeriodsOverlap() ... etc...

Stephek (changes requested):

Stephek (approved):

PR #563, opened 2022-12-06
EI-2003 allow multiple employment check refresh service requests

David-B-Read:


src/SFA.DAS.EmployerIncentives.Api/Controllers/JobCommandController.cs
```diff
@@ -64,24 +64,24 @@ private static RefreshLegalEntitiesCommand ParseJobRequestToRefreshCommand(JobRe
...
-                            checkTypeRequest.ToString(), 
-                            application.AccountLegalEntityId, 
-                            application.ULN, 
-                            serviceRequest.TaskId, 
-                            serviceRequest.DecisionReference, 
-                            serviceRequest.TaskCreatedDate)
+                            employmentCheckRefreshRequest.CheckType,
+                            application.AccountLegalEntityId,
+                            application.ULN,
+                            employmentCheckRefreshRequest.ServiceRequest.TaskId,
```
Stephek, line 80: If an application has a ServiceRequest then that should override the ServiceRequest of the request.

Stephek (changes requested):

Stephek (approved):

PR #584, opened 2023-07-26
EI-2073 check for ApprenticeshipIncentive records before deleting account

David-B-Read:


src/SFA.DAS.EmployerIncentives.Commands/Persistence/ApprenticeshipIncentiveDomainRepository.cs
```diff
@@ -110,6 +110,13 @@ public async Task<Domain.ApprenticeshipIncentives.ApprenticeshipIncentive> FindB
             return _apprenticeshipIncentiveFactory.GetExisting(incentive.Id, incentive);
         }
 
+        public async Task<List<Domain.ApprenticeshipIncentives.ApprenticeshipIncentive>> FindByAccountLegalEntity(long accountLegalEntityId)
```
LiamWright, line 113: Does this need all the preceding namespace declarations? Doesn't this normally happen when two namespaces are the same as the class name?

LiamWright (commented):
Minor comment, and its more of a question than anything else

owencampbell (approved):

src/SFA.DAS.EmployerIncentives.Commands/Persistence/ApprenticeshipIncentiveDomainRepository.cs
David-B-Read: It's because ApprenticeshipIncentive is also defined as a namespace - SFA.DAS.EmployerIncentives.Commands.ApprenticeshipIncentive as introduced by this project - so the namespace prefix has to be explicitly defined

